PR #119, opened 2022-11-30
Added support for closing AvMenuTreeViewWindow automatically when it loses focus.

raiti-chan:
I felt that this type of selection window should close automatically when it loses focus.
(For example, a selection window for an ObjectField, etc.).
Now I could make it close automatically when it loses focus, but I think I could also get closer by using `ShowAuxWindow`.


bdunderscore:
Hmm. I checked out your PR locally but it does not seem to actually receive the OnLostFocus event when the window loses focus for some reason.

PR #130, opened 2022-12-02
Add validation for menu icons.

raiti-chan:
The VRCSDK has a size limit and a compression setting limit for MenuIcon.
If there are icons in the MenuInstaller menu that exceed these limits, they are uploaded as they are.
The upload itself is normal, and there is no problem with VRC operation, but at least an error occurs in the performance window, so it would be better to at least send an error to MenuInstaller.

Depending on the situation, it may be necessary to remove the icon or make an error when building, but since the menu settings themselves are the responsibility of the asset creator, it may be better to simply warn the author so as not to cause confusion among asset users.


bdunderscore:
ユーザーへ対処できないエラーを投げるような仕組みは避けたいので、できれば問題となるアイコンをビルド時に削除したいけど、それはそれで別問題なのでissueを作ってとりあえずマージします。ありがとうございます！

PR #179, opened 2023-01-11
feat: add simplified chinese translation

gizmo-ds:
Add Simplified Chinese translation


gizmo-ds:
done.
@bdunderscore PTAL

bdunderscore:
Hm. This leaves open the question of what to fall back to when I add new strings in the future. From a Chinese user's perspective, would it be preferable to show the English or Japanese strings in that case?

gizmo-ds:
> Hm. This leaves open the question of what to fall back to when I add new strings in the future. From a Chinese user's perspective, would it be preferable to show the English or Japanese strings in that case?

In my opinion, fallback to English is a good idea.

PR #193, opened 2023-01-28
Add MA Icon & Assign to MA Component Scripts

rrazgriz:
Small change that adds the MA logo to user-facing component scripts. This makes it easier to see when using enhanced hierarchy tools like QHierarchy and Hierarchy2, and improves look & feel in general.

I used the "M" ears cropped from the repository's banner image, rescaled to 32x32 with increased alpha for visibility. It is visible in both dark and light editor themes. I tried the other part of the logo (the "box of components") and putting the M on a script icon, but they seemed too cluttered. 

pfuuOPGP2u
em6QCy1cE6


bdunderscore:
I considered doing this, however it results in an extremely cluttered view when gizmos are turned on, and I expect a lot of users won’t know how to turn off the 3D icons in question. Do you know of a way to disable those icons by default?

rrazgriz:
Good point - I have 3D icons off by default so they just appeared as small icons in scene view for me.

[`GizmoUtility.SetIconEnabled()`](https://docs.unity3d.com/2022.2/Documentation/ScriptReference/GizmoUtility.SetIconEnabled.html) is perfect for this and exists in Unity 2022 and up, but not in earlier versions. 

In earlier versions, this functionality is buried in [`AnnotationUtility.bindings.cs`](https://github.com/Unity-Technologies/UnityCsReference/blob/2019.4/Editor/Mono/Annotation/AnnotationUtility.bindings.cs), which could be accessed using reflection, though it may be a bit awkward. These bindings appear to have not changed between 2019.4 and 2022.1, so the reflection interface wouldn't be too brittle.

What do you think?

rrazgriz:
Looked a bit more and Acegikmo provides a utility function for this [here](http://answers.unity.com/answers/1722605/view.html). This could be implemented as:

```cs
static MethodInfo setIconEnabled;
static MethodInfo SetIconEnabled => setIconEnabled = setIconEnabled ?? Assembly.GetAssembly(typeof(Editor))?.GetType("UnityEditor.AnnotationUtility")?.GetMethod("SetIconEnabled", BindingFlags.Static | BindingFlags.NonPublic);
public static void SetGizmoIconEnabled(Type type, bool on)
{
    if (SetIconEnabled == null) return;
    const int MONO_BEHAVIOR_CLASS_ID = 114; // https://docs.unity3d.com/Manual/ClassIDReference.html
    SetIconEnabled.Invoke(null, new object[] { MONO_BEHAVIOR_CLASS_ID, type.Name, on ? 1 : 0 });
}

public static void DisableGizmoIcons()
{
    SetGizmoIconEnabled(typeof(ModularAvatarBoneProxy), false);
    SetGizmoIconEnabled(typeof(ModularAvatarBlendshapeSync), false);
    SetGizmoIconEnabled(typeof(ModularAvatarMenuInstaller), false);
    SetGizmoIconEnabled(typeof(ModularAvatarMergeAnimator), false);
    SetGizmoIconEnabled(typeof(ModularAvatarMergeArmature), false);
    SetGizmoIconEnabled(typeof(ModularAvatarParameters), false);
    SetGizmoIconEnabled(typeof(ModularAvatarPBBlocker), false);
    SetGizmoIconEnabled(typeof(ModularAvatarVisibleHeadAccessory), false);
}

bdunderscore:
Yeah, having something like that as a one-time initialization would be perfect!

rrazgriz:
Added functionality to use reflection to disable these icons. I put in a note about `GizmoUtility` in Unity 2022+ but didn't include the conditional define logic, as it adds overhead to testing, but it seems unlikely that this utility will be used with 2022 for a while.

It uses `SessionState` to check if it's already run this session, so someone could enable the icons on a per-session basis if they wanted, without them resetting every script reload. I couldn't find a way to only run once on a per-project basis without making a file or something or using `EditorPrefs`, which is system-wide.

bdunderscore:
Thanks! Let me just check with the logo artist before merging - the logo itself isn't under the MIT license and has limitations on modifications (see COPYING.md).

bdunderscore:
I've got the OK from Pumo, so I'm going to go ahead and merge this - however, I'm going to move the actual icon out into the images repository to keep the licensing situation a bit clearer. Thanks for the contribution!

PR #216, opened 2023-02-18
fix: sometimes the MA icons aren't disabled properly

bdunderscore:
It seems that sometimes static initializers are created before the icon annotations are updated. This change checks for this case, and defers the icon-disabling until after the necessary annotation records are created.


bdunderscore:
Tagging @rrazgriz , does this look correct to you?

rrazgriz:
Unity Editor static initializer trouble, always a pain.

Fix looks good to me, I gave it a test and it works in my project upon restart.

bdunderscore:
This was tricky to repro as it only is an issue immediately after upgrading - on the next session restart, the annotations will have been created and so things will work.

PR #237, opened 2023-03-13
add some qol

float3:


Packages/nadena.dev.modular-avatar/Editor/AvatarProcessor.cs
```diff
@@ -102,6 +114,8 @@ private static void ApplyToCurrentAvatar()
                 BuildReport.Clear();
 
                 ProcessAvatar(avatar);
+                avatar.SetActive(false);
```
bdunderscore, line 117: Hmm. Disabling the original avatar is good for some workflows, but gets in the way in others (in particular I'll often bake and rebake the same avatar a lot of times while debugging 😅 ). I think I'd prefer to leave this step out?

float3: Does disabling prevent that though? you can bake a disabled Avatar iirc. if you still don't want this I'll remove that step.

bdunderscore: I guess I'm wondering why disabling is needed; we already shift the position of the avatar so you can visually check that things look right.

float3: https://feedback.vrchat.com/sdk-bug-reports/p/sdk-disables-all-publish-buttons-if-just-one-avatar-in-scene-has-an-error

float3: While MA's components might cause this other tools' components do

float3: I wonder why MA's components don't cause this bug?

PR #241, opened 2023-03-19
fix: retarget is not performed if rootBone is the only bone to be retargeted

### LINKED ISSUE #240: rootBone override is not retargeted
m_RootBone is not retargeted (if no bones are set to the SkinnedMeshRenderer).

This is general case for AvatarOptimizer.

anatawa12:
Closes #240 

This also fixes bone retarget is not performed if mesh is none


Packages/nadena.dev.modular-avatar/Editor/MeshRetargeter.cs
```diff
@@ -165,8 +167,11 @@ public Mesh Retarget()
             avatarTransform.localScale = Vector3.one;
 
             src = renderer.sharedMesh;
-            dst = Mesh.Instantiate(src);
-            dst.name = "RETARGETED: " + src.name;
+            if (src != null)
```
bdunderscore, line 170: If src is null, why not just return immediately? The renderer is going to be broken with no mesh anyway.

bdunderscore:
Also, please include unit tests for this bugfix (as-is I'm not sure what conditions will trigger the bug, or what the expected behavior is, a unit test would help clarify that)

Packages/nadena.dev.modular-avatar/Editor/MeshRetargeter.cs
anatawa12: AvatarOptimizer will assign mesh later so I want MA to retarget bones (especially root transform & bounding box) of SkinnedMeshRenderer with null mesh.

PR #265, opened 2023-04-04
feat: add Foot=Ankle mapping, more hand bone names

enitama:
This adds support for Kokuri Ruru, which uses:

- `Ankle_L`/`Ankle_R` for foot bones.
- Hand bone names generated with the regex replacement `"((?!Thunb)[A-Za-z]+?)([1-3]_[LR])"` → `"$1$2", "$1Finger$2"`.
  - _Note:_ This is overspecified. Kokuri Ruru's thumbs use `Thumb[1-3]_L` not `ThumbFinger[1-3]_L` unlike the other fingers, however these extra bones are left in to account for the possibility of other avatars using such bone names. Happy to remove if considered unnecessary.

Result:

スクリーンショット 2023-04-04 223135


enitama:
Sorry for the noisy commit log. I'm relatively unfamiliar with Conventional Commits and wasn't sure what the right naming should be.

I went with 549ce8f0d3d6ef7c13c045c86398b61efc4db7d1 ("fix: add more heuristic bone mappings") first while retaining avatar names, before noticing that following the newer 319fd9a2ff6cb83c45b3d78771378ef4ae233373 ("feat: add x bone mapping") and using a generic commit message would probably work better.

Avatar names are currently still included in the commit desecription. Happy to remove if unnecessary.

bdunderscore:
"Index_Distal_L" and friends should be redundant per the name normalization logic here (the presence or absence of spaces, periods, and underscores is ignored): https://github.com/bdunderscore/modular-avatar/blob/main/Packages/nadena.dev.modular-avatar/Editor/HeuristicBoneMapper.cs#L75-L81

bdunderscore:
Thanks for the contribution!

PR #270, opened 2023-04-07
Modular Avatar Outfit Compatibility and Wardrobe System Integration

HotKonti:
This pull request introduces new functionality to enable compatibility with outfits sold on platforms like Booth, utilizing the modular-avatar system and integrating them with the wardrobe system. 

### Changes
- Added the `ModulerAvatarParameterRevalue` component to handle collisions with pre-defined external float or int parameters within the module.
- Extended the functionality of the `AvatarProcessor` to support this new feature.

### Benefits
With these changes, users can now easily change and manage outfits through the wardrobe system, seamlessly integrating purchased outfits into their avatars.


HotKonti:
I apologize. I would like to write test code for the project, but I'm not sure how to test it after writing the test code. Could you please explain the process so that I can write the test code and add the commit? Thank you.

bdunderscore:
What is this wardrobe system you refer to? And, can you provide a brief summary of the intended behavior of this new component?

HotKonti:
image

image

image

This feature allows for easy adjustment of the menu's parameter values, the animator's transition condition parameter values, and the blend tree's threshold values when creating a wardrobe similar to the one described above. This makes it possible for clothing buyers to use various options provided in the wardrobe and change the design of the chosen clothes without having to perform any special tasks.

bdunderscore:
Thanks for the explanation. The approach proposed in this PR has some performance issues as it'll require a separate layer (and two always-active transitions) for each outfit, so I'm not inclined to adopt it as proposed.  I currently plan to support clothing changes using the "action toggle object" component that's currently on the dev branch; I'm still working on the UI/UX a bit, but you can see the draft documentation at https://modular-avatar.nadena.dev/dev/docs/tutorials/menu/

HotKonti:
Thank you for your input regarding performance issues. We appreciate your suggestion and will work on designing and developing a feature that merges layers for the same type of state transition, taking into consideration whether it is a write default or not. We will then resubmit the pull request for your review.

bdunderscore:
I would suggest opening an issue with the proposed semantics of your solution first, so we can go over whether it'll fit in with the future plans of MA before you spend too much time implementing it - as I mentioned, I have some code that's cooking over on the dev branch that implements a different strategy for this sort of thing.

bdunderscore:
I'm going to close this as I'm still exploring how the UI/UX of animation generation should work. In the meantime, I'll look into exposing hooks to allow you to plug into the MA framework more easily.

PR #284, opened 2023-04-25
Add object remover component

GotoFinal:
A simple component that you can put on any object to just remove it during the build, additionaly a button can be used to hide it from hierarchy view too (it's not removed, just hidden, right click on parent object can be used to restore it) making your avatar setup much more clean if you often only use parts of some avatars/clothing.

image
image

Comparsion on my sample avatar before and after adding this component: 
image
image

### DIFF
--- a/Packages/nadena.dev.modular-avatar/Editor/AvatarProcessor.cs
+++ b/Packages/nadena.dev.modular-avatar/Editor/AvatarProcessor.cs
@@ -191,6 +191,7 @@ namespace nadena.dev.modular_avatar.core.editor
 
                         var context = new BuildContext(vrcAvatarDescriptor);
 
+                        ObjectRemoverProcessor.OnPreprocessAvatar(avatarGameObject);
                         new RenameParametersHook().OnPreprocessAvatar(avatarGameObject, context);
                         new MergeAnimatorProcessor().OnPreprocessAvatar(avatarGameObject, context);
                         context.AnimationDatabase.Bootstrap(vrcAvatarDescriptor);
@@ -201,7 +202,6 @@ namespace nadena.dev.modular_avatar.core.editor
                         new VisibleHeadAccessoryProcessor(vrcAvatarDescriptor).Process(context);
                         new RemapAnimationPass(vrcAvatarDescriptor).Process(context.AnimationDatabase);
                         new BlendshapeSyncAnimationProcessor().OnPreprocessAvatar(avatarGameObject, context);
-                        ObjectRemoverProcessor.OnPreprocessAvatar(avatarGameObject);
                         PhysboneBlockerPass.Process(avatarGameObject);
 
                         AfterProcessing?.Invoke(avatarGameObject, context);
--- a/Packages/nadena.dev.modular-avatar/Editor/AvatarProcessor.cs
+++ b/Packages/nadena.dev.modular-avatar/Editor/AvatarProcessor.cs
@@ -201,6 +201,7 @@ namespace nadena.dev.modular_avatar.core.editor
                         new VisibleHeadAccessoryProcessor(vrcAvatarDescriptor).Process(context);
                         new RemapAnimationPass(vrcAvatarDescriptor).Process(context.AnimationDatabase);
                         new BlendshapeSyncAnimationProcessor().OnPreprocessAvatar(avatarGameObject, context);
+                        ObjectRemoverProcessor.OnPreprocessAvatar(avatarGameObject);
                         PhysboneBlockerPass.Process(avatarGameObject);
 
                         AfterProcessing?.Invoke(avatarGameObject, context);
--- a/Packages/nadena.dev.modular-avatar/Editor/ObjectRemoverProcessor.cs
+++ b/Packages/nadena.dev.modular-avatar/Editor/ObjectRemoverProcessor.cs
@@ -1,0 +1,30 @@
+﻿using UnityEditor;
+using UnityEngine;
+
+namespace nadena.dev.modular_avatar.core.editor
+{
+    public static class ObjectRemoverProcessor
+    {
+        internal static void OnPreprocessAvatar(GameObject avatarGameObject)
+        {
+            var toRemove = avatarGameObject.transform.GetComponentsInChildren<ModularAvatarObjectRemover>(true);
+
+            foreach (var remove in toRemove)
+            {
+                Object.DestroyImmediate(remove.gameObject);
+            }
+        }
+
+
+        [MenuItem("GameObject/ModularAvatar/Restore Removed", false, 49)]
+        static void RestoreRemoved(MenuCommand menuCommand)
+        {
+            if (!(menuCommand.context is GameObject obj)) return;
+            foreach (Transform child in obj.transform)
+            {
+                var toRemove = child.GetComponent<ModularAvatarObjectRemover>();
+                if (toRemove != null) toRemove.Restore();
+            }
+        }
+    }
+}
--- a/Packages/nadena.dev.modular-avatar/Editor/ObjectRemoverProcessor.cs
+++ b/Packages/nadena.dev.modular-avatar/Editor/ObjectRemoverProcessor.cs
@@ -1,5 +1,4 @@
-﻿using UnityEditor;
-using UnityEngine;
+﻿using UnityEngine;
 
 namespace nadena.dev.modular_avatar.core.editor
 {
@@ -7,23 +6,10 @@ namespace nadena.dev.modular_avatar.core.editor
     {
         internal static void OnPreprocessAvatar(GameObject avatarGameObject)
         {
-            var toRemove = avatarGameObject.transform.GetComponentsInChildren<ModularAvatarObjectRemover>(true);
-
-            foreach (var remove in toRemove)
+            var removers = avatarGameObject.transform.GetComponentsInChildren<ModularAvatarObjectRemover>(true);
+            foreach (var remover in removers)
             {
-                Object.DestroyImmediate(remove.gameObject);
-            }
-        }
-
-
-        [MenuItem("GameObject/ModularAvatar/Restore Removed", false, 49)]
-        static void RestoreRemoved(MenuCommand menuCommand)
-        {
-            if (!(menuCommand.context is GameObject obj)) return;
-            foreach (Transform child in obj.transform)
-            {
-                var toRemove = child.GetComponent<ModularAvatarObjectRemover>();
-                if (toRemove != null) toRemove.Restore();
+                foreach (var toRemove in remover.objectsToRemove) Object.DestroyImmediate(toRemove);
             }
         }
     }
--- a/Packages/nadena.dev.modular-avatar/Editor/ObjectRemoverProcessor.cs
+++ b/Packages/nadena.dev.modular-avatar/Editor/ObjectRemoverProcessor.cs
@@ -1,4 +1,5 @@
-﻿using UnityEngine;
+﻿using System.Collections.Generic;
+using UnityEngine;
 
 namespace nadena.dev.modular_avatar.core.editor
 {
@@ -7,10 +8,13 @@ namespace nadena.dev.modular_avatar.core.editor
         internal static void OnPreprocessAvatar(GameObject avatarGameObject)
         {
             var removers = avatarGameObject.transform.GetComponentsInChildren<ModularAvatarObjectRemover>(true);
+            var objectsToRemove = new List<GameObject>();
             foreach (var remover in removers)
             {
-                foreach (var toRemove in remover.objectsToRemove) Object.DestroyImmediate(toRemove);
+                objectsToRemove.AddRange(remover.objectsToRemove);
             }
+
+            foreach (var toRemove in objectsToRemove) Object.DestroyImmediate(toRemove);
         }
     }
 }
--- a/Packages/nadena.dev.modular-avatar/Runtime/ModularAvatarObjectRemover.cs
+++ b/Packages/nadena.dev.modular-avatar/Runtime/ModularAvatarObjectRemover.cs
@@ -1,0 +1,52 @@
+﻿using UnityEditor;
+using UnityEngine;
+
+namespace nadena.dev.modular_avatar.core
+{
+    [ExecuteInEditMode]
+    [DisallowMultipleComponent]
+    [AddComponentMenu("Modular Avatar/MA Object Remover")]
+    public class ModularAvatarObjectRemover : AvatarTagComponent
+    {
+        public bool hideInHierarchy;
+
+        public ModularAvatarObjectRemover()
+        {
+            EditorApplication.delayCall += Refresh;
+        }
+
+        private void OnEnable()
+        {
+            Refresh();
+        }
+
+        private void Refresh()
+        {
+            if (Application.isPlaying) return;
+            gameObject.SetActive(false);
+            if (hideInHierarchy) Hide();
+            // components can be always restored from prefabs, so removing them now does not break anything and will
+            // provide better compatibility with other tools
+            foreach (var componentsInChild in gameObject.GetComponentsInChildren<MeshRenderer>())
+            {
+                if (PrefabUtility.IsPartOfAnyPrefab(componentsInChild)) DestroyImmediate(componentsInChild);
+            }
+            foreach (var componentsInChild in gameObject.GetComponentsInChildren<SkinnedMeshRenderer>())
+            {
+                if (PrefabUtility.IsPartOfAnyPrefab(componentsInChild)) DestroyImmediate(componentsInChild);
+            }
+        }
+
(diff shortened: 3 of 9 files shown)


Packages/nadena.dev.modular-avatar/Runtime/ModularAvatarObjectRemover.cs
```diff
@@ -0,0 +1,52 @@
...
+            // components can be always restored from prefabs, so removing them now does not break anything and will
+            // provide better compatibility with other tools
+            foreach (var componentsInChild in gameObject.GetComponentsInChildren<MeshRenderer>())
+            {
+                if (PrefabUtility.IsPartOfAnyPrefab(componentsInChild)) DestroyImmediate(componentsInChild);
+            }
+            foreach (var componentsInChild in gameObject.GetComponentsInChildren<SkinnedMeshRenderer>())
+            {
+                if (PrefabUtility.IsPartOfAnyPrefab(componentsInChild)) DestroyImmediate(componentsInChild);
+            }
```
bdunderscore, line 37: This feels like surprising behavior from a user perspective - when I hide an object and restore it later, it's missing a bunch of MeshRenderers, and any prefab overrides are lost.

Packages/nadena.dev.modular-avatar/Runtime/ModularAvatarObjectRemover.cs
```diff
@@ -0,0 +1,52 @@
...
+namespace nadena.dev.modular_avatar.core
+{
+    [ExecuteInEditMode]
+    [DisallowMultipleComponent]
+    [AddComponentMenu("Modular Avatar/MA Object Remover")]
+    public class ModularAvatarObjectRemover : AvatarTagComponent
+    {
+        public bool hideInHierarchy;
+
+        public ModularAvatarObjectRemover()
```
bdunderscore, line 13: This should probably be in OnValidate or Awake

Packages/nadena.dev.modular-avatar/Runtime/ModularAvatarObjectRemover.cs
```diff
@@ -0,0 +1,52 @@
...
+            foreach (var componentsInChild in gameObject.GetComponentsInChildren<SkinnedMeshRenderer>())
+            {
+                if (PrefabUtility.IsPartOfAnyPrefab(componentsInChild)) DestroyImmediate(componentsInChild);
+            }
+        }
+
+        public void Hide()
+        {
+            hideInHierarchy = true;
+            gameObject.hideFlags |= HideFlags.HideInHierarchy;
```
bdunderscore, line 43: When you make changes to a prefab instance, it's important to call `PrefabUtility.RecordPrefabInstancePropertyModifications` (and `EditorUtility.SetDirty`, in case it's not part of a prefab instance) to ensure the changes are saved properly.

Packages/nadena.dev.modular-avatar/Editor/ObjectRemoverProcessor.cs
```diff
@@ -0,0 +1,30 @@
...
+            var toRemove = avatarGameObject.transform.GetComponentsInChildren<ModularAvatarObjectRemover>(true);
+
+            foreach (var remove in toRemove)
+            {
+                Object.DestroyImmediate(remove.gameObject);
+            }
+        }
+
+
+        [MenuItem("GameObject/ModularAvatar/Restore Removed", false, 49)]
```
bdunderscore, line 19: There should probably be a validate function here to disable this when there are no hidden/removed objects.

Packages/nadena.dev.modular-avatar/Runtime/ModularAvatarObjectRemover.cs
bdunderscore: Would it be better to just set EditorOnly (in addition to hiding from view?)

Packages/nadena.dev.modular-avatar/Editor/AvatarProcessor.cs
```diff
@@ -184,6 +184,7 @@ public static void ProcessAvatar(GameObject avatarGameObject)
                     new VisibleHeadAccessoryProcessor(vrcAvatarDescriptor).Process(context);
                     new RemapAnimationPass(vrcAvatarDescriptor).Process(context.AnimationDatabase);
                     new BlendshapeSyncAnimationProcessor().OnPreprocessAvatar(avatarGameObject, context);
+                    ObjectRemoverProcessor.OnPreprocessAvatar(avatarGameObject);
```
bdunderscore, line 187: This is a rather late phase of processing to remove objects at - it means that any menu installers, etc will still be applied.

bdunderscore:
At a high level - this seems like a useful utility, though I'm not entirely sure it belongs in the MA umbrella (and can be implemented using EditorOnly + hierarchy hiding without needing a dedicated hook). Some high level thoughts on UX:

1. One issue with this is that there's no easy way to find which objects in your hierarchy have hidden objects under them. Have you given any thought to e.g. having an option to show all hidden objects, or to highlight parents of hidden objects in the hierarchy?
2. If I want to restore just a single hidden object from a parent object, but there are multiple hidden objects under the same parent, how do I make sure I re-hide all the others and don't forget any?
3. For new MA components, unit tests and documentation are required.

Packages/nadena.dev.modular-avatar/Runtime/ModularAvatarObjectRemover.cs
GotoFinal: Well my assumption is that this component is like removing but sadly we just can't remove objects without breaking a prefab. The component is technically still there as removing component from a prefab leaves you with marker that it got removed
While I do prefer this behaviour, I'm fine to just remove it, just some tools then keep complaining about these mesh renderes or want to use them for their actions. 
From my manual testing removing component that is part of a prefab does NOT remove any data, even the changes you did inside the component are still preserved: 
image
image
And its back to:
image

Packages/nadena.dev.modular-avatar/Runtime/ModularAvatarObjectRemover.cs
GotoFinal: This is a workaround for unity issue sadly, I need this to get called even if the object is disabled, because unity strips HideFlags from prefab instances on save, and thats the only way I found to restore them

Packages/nadena.dev.modular-avatar/Runtime/ModularAvatarObjectRemover.cs
GotoFinal: This changes cannot be saved to a prefab anyways: https://forum.unity.com/threads/prefab-sub-asset-hideflags-dont-persist.1324920/

GotoFinal:
@bdunderscore 
> 1. One issue with this is that there's no easy way to find which objects in your hierarchy have hidden objects under them. Have you given any thought to e.g. having an option to show all hidden objects, or to highlight parents of hidden objects in the hierarchy?

I did not find any good way to do this, the idea is that as a user I just want to remove an object but sadly I can't because its part of a prefab, so the chances I want it back are rather slim. 
Maybe adding a global action to unhide all would be good idea, but overall I don't want to clutter the user with informations about stuff they want gone.

> If I want to restore just a single hidden object from a parent object, but there are multiple hidden objects under the same parent, how do I make sure I re-hide all the others and don't forget any?

You don't need to, hidding and removing is separate action, objects will be still removed on build, hidding is just an esthetic option to help you keep a clean look of a project and remove visual clutter. But you can just look for any object that have the component and use the button to hide them again. 

One other option that maybe would improve both of these issues is to make object remover work on parent object directly, instead of adding it to the object you want to remove, you put it on parent and have a list of objects to hide? I think this would improve both of the points you make here. 


> this seems like a useful utility, though I'm not entirely sure it belongs in the MA umbrella (and can be implemented using EditorOnly
Well not fully, as I still want the objects to be gone from my uploaded avatar, so I need something to actually remove them before uploading. This can actually be useful in some limited cases, from just marginally better performance of having less objects to having some physic bones on object chains and not dealing with issue of multiple children

Packages/nadena.dev.modular-avatar/Runtime/ModularAvatarObjectRemover.cs
bdunderscore: Ah, okay. Please add a comment describing the need for this workaround in this case.

bdunderscore:
Hiding and removing being separate makes a lot of sense - perhaps, then, there should be a menu option under `Tools -> Modular Avatar` to show all hidden objects, _without removing the hide flag_? That way, if you want to unhide an object, you temporarily show all hidden objects, then clear the flag, and then un-show hidden objects.

I do think I would prefer not removing the meshrenderers implicitly - the user can always do this themselves, and having that happen magically doesn't really fit with the nondestructive policy of Modular Avatar.

GotoFinal:
@bdunderscore I made removing a separate toggle, but also I want to once again remind its not actually destructive as I only remove prefab instance components, they still exist and save all of the data. 

Overall I decided to try a bit different approach, let me tell what you think about it and if I should then continue it in this way.
The component now is not marking objects to remove, but contains a list of objects to remove - this solves issues of hiding and unhiding components and allows to group stuff together. 

I think video should show it well:

https://user-images.githubusercontent.com/6942692/235809674-f7067e95-6aa7-4273-bfd6-46ebca72cf25.mp4

Packages/nadena.dev.modular-avatar/Editor/ObjectRemoverProcessor.cs
GotoFinal: no longer needed

Packages/nadena.dev.modular-avatar/Editor/AvatarProcessor.cs
GotoFinal: moved to the top as this is actually valid use, people could create multiple versions of same avatar by just creating prefab variants and removing certain elements including the modular avatar components

Packages/nadena.dev.modular-avatar/Runtime/ModularAvatarObjectRemover.cs
GotoFinal: Added option to restore and made this behaviour use a separate toggle

bdunderscore:
I just got back from a trip and haven't had a chance to look at the diff yet - but looking at the video I think I like this approach. That said, the one thing I'd be worried about is the possibility of the remover component itself becoming inaccessible due to loops. For example, if you were to have an object at `x/y/z` and add an Object Remover which references `x`, you won't be able to restore it. There would still need to be an escape hatch for this kind of situation.

GotoFinal:
Good catch, I can just add a rule to prevent adding itself (or parent) to the list or at least not hide it then - nothing more complicated seems worth it

bdunderscore:
I do wonder if there are scenarios where you could have a loop between two objects - perhaps as a result of prefab overrides being applied or cleared? It might be more reliable to just have a way to recover if it happens.

GotoFinal:
> I do wonder if there are scenarios where you could have a loop between two objects - perhaps as a result of prefab overrides being applied or cleared? It might be more reliable to just have a way to recover if it happens.
With auto disabling enabled the loop would break after single scene reload as its not intended use, but otherwsie...
I can just make it not hide objects with this component, then I don't think it can loop in any way and adding such way would be another menu to handle for a case that does not exist to be added to translation files, tests, documentation. I want to keep it simple. And I don't see any use case to want to hide such objects.

bdunderscore:
I'm wondering about more this case:


```
Root
 +- A
 |   \- B
 \- C
     \- D
```

Where D hides A and B hides C - there's no way to break this loop currently, right?

GotoFinal:
> Where D hides A and B hides C - there's no way to break this loop currently, right?

I pushed the change i was talking about, now this can't happen as the component will just not hide any object that contains other removers to prevent any issues like that - especially that its not intended use of hiding logic, you only want to hide clutter you are not doing anything with

Packages/nadena.dev.modular-avatar/Runtime/ModularAvatarObjectRemover.cs
```diff
@@ -0,0 +1,133 @@
...
+{
+    [ExecuteInEditMode]
+    [AddComponentMenu("Modular Avatar/MA Object Remover")]
+    public class ModularAvatarObjectRemover : AvatarTagComponent
+    {
+        public bool hideInHierarchy = true;
+        public bool keepDisabled = true;
+        public bool removePrefabComponents;
+        public List<GameObject> objectsToRemove = new List<GameObject>();
+        private List<GameObject> _lastObjectsToRemove;
```
bdunderscore, line 17: Should this be serialized, so that changes that happen outside of the inspector are reverted properly?

Packages/nadena.dev.modular-avatar/Runtime/ModularAvatarObjectRemover.cs
```diff
@@ -0,0 +1,133 @@
...
+            if (Application.isPlaying) return;
+            hideInHierarchy = false;
+            removePrefabComponents = false;
+            Unhide();
+            RestorePrefabComponents();
+            objectsToRemove.Clear();
+            _lastObjectsToRemove = null;
+        }
+
+        public void OnListChange()
```
bdunderscore, line 36: Should we trigger a check for changes from OnValidate as well? (again, to try to catch changes made from editor scripts/prefab override applications/reverts/etc)

Packages/nadena.dev.modular-avatar/Runtime/ModularAvatarObjectRemover.cs
```diff
@@ -0,0 +1,133 @@
...
+        public void RemovePrefabComponents()
+        {
+            var toDelete = new List<Component>();
+
+            foreach (var toRemove in objectsToRemove)
+            {
+                if (removePrefabComponents) CollectComponentsToRemove(toDelete, toRemove.transform);
+            }
+
+            // NOTE: this is non-destructive, all components from prefabs can be restored with no data loss
```
bdunderscore, line 93: We do lose information on any components that were already removed on the prefab instance before we applied this change. This in turn means that undo won't work properly, as we don't record these removals in the undo record and don't retain the information needed to _not_ restore components that were already removed.

Packages/nadena.dev.modular-avatar/Editor/ObjectRemoverProcessor.cs
```diff
@@ -0,0 +1,16 @@
...
+namespace nadena.dev.modular_avatar.core.editor
+{
+    public static class ObjectRemoverProcessor
+    {
+        internal static void OnPreprocessAvatar(GameObject avatarGameObject)
+        {
+            var removers = avatarGameObject.transform.GetComponentsInChildren<ModularAvatarObjectRemover>(true);
+            foreach (var remover in removers)
+            {
+                foreach (var toRemove in remover.objectsToRemove) Object.DestroyImmediate(toRemove);
```
bdunderscore, line 12: This could throw a null reference exception if a MAOR is removed by a prior MAOR.

bdunderscore (changes requested):

Packages/nadena.dev.modular-avatar/Runtime/ModularAvatarObjectRemover.cs
GotoFinal: ONValidate then should still handle it

Packages/nadena.dev.modular-avatar/Runtime/ModularAvatarObjectRemover.cs
GotoFinal: Can't be really triggered fro OnValidate, unless with delayCall.
But at this point I think its all out of scope of this idea, whole idea is to have some easy way of getting rid of parts of prefab you dont want to use without need of breaking the prefab and ability to restore it in case you change mind. Allowing for quicker creation, better optimisation and cleaner project, so I don't see any use cases for such situations

Packages/nadena.dev.modular-avatar/Runtime/ModularAvatarObjectRemover.cs
GotoFinal: Kinda same as above, at this point I think its all out of scope of this idea, whole idea is to have some easy way of getting rid of parts of prefab you dont want to use without need of breaking the prefab and ability to restore it in case you change mind. Allowing for quicker creation, better optimisation and cleaner project, so I don't see any use cases for such situations.
For needs of this repo I guess removing feature can be removed, as its not too important anyways, but the idea was to just have nice way to automate removing all the components from all the stuff you don't use, so if restoring will restore too much its not a big deal compared to saving time on removing and searching for it manually without breaking anything

bdunderscore:
No response on this and #307, closing this for now.

PR #290, opened 2023-04-28
CRITICAL: Fixed missing namespaces that are required by SDK 3.2.x

BluWizard10:
Modular Avatar will cease to work with the new SDK 3.2.0 due to changes to the Extension Methods being moved to the VRC.Core namespace. So a critical change has been made in order to make sure it works again.

I have went ahead and added them as part of this Pull Request. **As this is backwards-compatible, it is highly advised that this gets merged ASAP so that Modular Avatar can function properly on SDK 3.2.0 once it's released to the Public.**

### DIFF
--- a/Packages/nadena.dev.modular-avatar/Editor/MergeArmatureHook.cs
+++ b/Packages/nadena.dev.modular-avatar/Editor/MergeArmatureHook.cs
@@ -33,6 +33,7 @@ using VRC.Dynamics;
 using VRC.SDK3.Avatars.Components;
 using VRC.SDK3.Dynamics.PhysBone.Components;
 using Object = UnityEngine.Object;
+using VRC.Core;
 
 namespace nadena.dev.modular_avatar.core.editor
 {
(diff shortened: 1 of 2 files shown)


bdunderscore:
A fix for this is already present on the `dev` branch and beta release. While the beta/dev branch isn't ready for release yet, I'm waiting to see if the VRCSDK will restore the old extension method location before cutting a patch release for this - [the relevant canny](https://feedback.vrchat.com/open-beta/p/sdk-320-beta1-extensionmethodsgetoraddcomponent-was-moved) was marked as "available in future release".

BluWizard10:
Yes, Momo said in the VRChat Discord that they would restore the old extension method. However, have made it clear that they intend to fully deprecate the old extension method and move it to the VRC.Core namespace permanently in the near future.

bdunderscore:
Makes sense. Since there's not an immediate rush, the transition to VRC.Core will happen when the menu generation feature is ready for release.

PR #307, opened 2023-05-20
Gc pass will now scan for unused bones and cleanup physbones and colliders

GotoFinal:
Currently GC only removes just very simple unreferenced objects, its usually very few objects so it does not help with optimisation or cleanup much. I wrote additional step of GC that collect all bones from all meshes and determines if they are still used or not so they can be correctly removed. Often when merging some avatars you can be left with huge amount of bones from their skirts/hairs if you decided to not use these parts at all. As long as you remove the mesh renderer from the component new GC script will detect this and remove the bones and related dynamic bones and colliders that no longer collide with anything.

### DIFF
--- a/Packages/nadena.dev.modular-avatar/Editor/OptimizationPasses/GCGameObjectsPass.cs
+++ b/Packages/nadena.dev.modular-avatar/Editor/OptimizationPasses/GCGameObjectsPass.cs
@@ -1,7 +1,9 @@
 ﻿using System;
 using System.Collections.Generic;
+using System.Linq;
 using UnityEditor;
 using UnityEngine;
+using VRC.Dynamics;
 using VRC.SDK3.Dynamics.PhysBone.Components;
 
 namespace nadena.dev.modular_avatar.core.editor
@@ -13,7 +15,13 @@ namespace nadena.dev.modular_avatar.core.editor
     {
         private readonly BuildContext _context;
         private readonly GameObject _root;
-        private readonly HashSet<GameObject> referencedGameObjects = new HashSet<GameObject>();
+        private readonly HashSet<GameObject> _referencedGameObjects = new HashSet<GameObject>();
+        private readonly HashSet<GameObject> _allBones = new HashSet<GameObject>();
+        private readonly HashSet<GameObject> _usedBones = new HashSet<GameObject>();
+        private readonly HashSet<VRCPhysBone> _physBones = new HashSet<VRCPhysBone>();
+        private readonly HashSet<VRCPhysBoneColliderBase> _physBoneColliders = new HashSet<VRCPhysBoneColliderBase>();
+        private int _removedObjects;
+        private int _removedComponents;
 
         internal GCGameObjectsPass(BuildContext context, GameObject root)
         {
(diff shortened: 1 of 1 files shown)


Packages/nadena.dev.modular-avatar/Editor/OptimizationPasses/GCGameObjectsPass.cs
```diff
@@ -23,11 +31,132 @@ internal GCGameObjectsPass(BuildContext context, GameObject root)
...
+        {
+            HashSet<int> usedBonesIndices = new HashSet<int>();
+            foreach (var boneWeight in renderer.sharedMesh.boneWeights)
+            {
+                // technically should add only if weight is > 0, but not worth the cost and does not have negative effects
+                usedBonesIndices.Add(boneWeight.boneIndex0);
+                usedBonesIndices.Add(boneWeight.boneIndex1);
+                usedBonesIndices.Add(boneWeight.boneIndex2);
+                usedBonesIndices.Add(boneWeight.boneIndex3);
+            }
```
bdunderscore, line 71: It's probably better to use https://docs.unity3d.com/ScriptReference/Mesh.GetAllBoneWeights.html - this not only deals with meshes with >4 weights (though I'm not sure if VRC support that anyway), but also results in cleaner code as you don't need to have this 4x add thing.

Packages/nadena.dev.modular-avatar/Editor/OptimizationPasses/GCGameObjectsPass.cs
```diff
@@ -23,11 +31,132 @@ internal GCGameObjectsPass(BuildContext context, GameObject root)
...
+            // remove phys bones that only affect bones that are not used
+            var bonesToRemove = new HashSet<VRCPhysBone>();
+            foreach (var vrcPhysBone in _physBones)
+            {
+                vrcPhysBone.InitTransforms();
+                if (vrcPhysBone.bones.Any(x => _usedBones.Contains(x.transform.gameObject))) continue;
+                // in case someone uses the component on chain of something that is not actually a bone
+                if (vrcPhysBone.bones.Any(x => !_allBones.Contains(x.transform.gameObject))) continue;
+                bonesToRemove.Add(vrcPhysBone);
+            }
```
bdunderscore, line 97: PhysBones can influence other things than just mesh renderers - lights, particle systems, audio sources, colliders, etc. We need to take this into account before deleting physbones here.

Packages/nadena.dev.modular-avatar/Editor/OptimizationPasses/GCGameObjectsPass.cs
```diff
@@ -23,11 +31,132 @@ internal GCGameObjectsPass(BuildContext context, GameObject root)
...
+                if (usedBone != null) _usedBones.Add(usedBone.gameObject);
+            }
+
+            foreach (var bone in bones)
+            {
+                if (bone != null) _allBones.Add(bone.gameObject);
+            }
+        }
+
+        private void Cleanup()
```
bdunderscore, line 86: This helper function needs a more specific name

bdunderscore (changes requested):
An interesting idea - there's bound to be a lot of avatar with e.g. leftover invisible skirt bone chains that could benefit from this.

That being said, optimizations like this can result in mysterious breakage, which is why I chose to start conservatively. In particular, I think this one now needs some unit tests to verify that it works properly.

Packages/nadena.dev.modular-avatar/Editor/OptimizationPasses/GCGameObjectsPass.cs
GotoFinal: Thats why there is second check for "all bones", if it contains anything that is not a bone it wlll be left alone.

bdunderscore: To clarify - if you have a bone which has, say, a light on it, and which is also in a renderer's bones array (but has no vertices weighted to it), then it will be in _allBones, but not in _usedBones. This in turn means a physbone that influences a bone like this can be dropped.

GotoFinal: in this case yes, but it must be actual bone, added in blender, and thats rather weird use case - not even use how it would work with a blender bone not weighted to anything, would probabvly require separate way to flag such weird cases to not be removed - as overall I want to remove objects like that as hard to tell apart bone that is no longer used because object is gone vs someone doing it on purpose. 
As if the skirt contained some additional components on it like lights or particle systems, they should be removed too.

bdunderscore:
I see you've added a WIP commit to this and to #284 - is that intended to be ready for review or is it still WIP?

bdunderscore:
Going to close this as there's been no response. Going forward, I expect deep optimizations to be more of something for https://github.com/anatawa12/AvatarOptimizer (the purpose of the GC pass here is to clean up after all the new empty game objects used for the menu system)

PR #324, opened 2023-06-05
Add the Mesh Settings component

bdunderscore:
This adds a Mesh Settings component which handles setting bounds, root bone, and probe anchor properties on child meshes.

This is based on prior work by @AoiKamishiro in #304 and @raiti-chan in #318 .

Closes: #134


bdunderscore:
TODO: Documentation pages (before merging)

Issue #478, opened 2023-10-08
feat: Non-VRChat Avatars support


kaikoga:
### Discussed in https://github.com/bdunderscore/modular-avatar/discussions/467

Wrap VRCSDK dependencies of Modular Avatars in compile flags to support non-VRChat avatars.

## Required Tasks

- [ ] Add `#if UNITY_2021_2_OR_NEWER`
- [ ] Add Version Define `MA_VRCSDK3_AVATARS` against "com.vrchat.avatars"
- [ ] `#if NDMF_VRCSDK3_AVATARS` VRChat specific MA components and passes
- [ ] Use ndmf 1.2.0 API to reduce VRCAvatarDescriptor usage
- [ ] `#if NDMF_VRCSDK3_AVATARS` VRCSDK component references within generic MA components

## Not right now (create another issue)

- UniVRM support: VRM0 / VRM1
  - Add Version Define `MA_VRM1` against "com.vrmc.vrm" and/or `MA_UNIVRM` against "com.vrmc.univrm"
  - Add VRM related implementations to MA, compile flags guarded
    - Add UniVRM component references to generic MA components
      - Ideas: Some PhysBones specific logics might also need to be able to handle SpringBones
    - Implement UniVRM specific MA components and passes , if MA is a better place
      - Ideas here
    - Other possible VRM specific optimization 
      - Ideas: VisibleHeadAccessory can make use of VRMFirstPerson
- Other non-VRChat avatar supports (by eager somebody)

PR #482, opened 2023-10-08
chore: use ndmf Avatar Root api where applicable

kaikoga:
Related issue: #478 

## What is this PR all about?

As ndmf 1.2.0 has implemented some cross platform Avatar root handling APIs in bdunderscore/ndmf#34, we can make use of them before introducing `#if MA_VRCSDK3_AVATARS` version define to minimize the big bang diff.

## Refactoring done:

- Utilize ndmf 1.2.0 API from in some of EditorContext and RuntimeUtil methods (The final commit is very aggressive!)
- Use them instead of `VRCAvatarDescriptor` everywhere
  - except VRC animation layer handlers and menu handlers, because they are very VRC specific
  - `VRCAvatarDescriptor` becomes either `BuildContext` or `Transform`, depending on context

This PR breaks a lot of internal APIs, which passes around `VRCAvatarDescriptor`, so this PR may be either directly merged or be used as a reference of what we should do and broken up into smaller PRs.

### DIFF
--- a/Runtime/RuntimeUtil.cs
+++ b/Runtime/RuntimeUtil.cs
@@ -63,28 +63,13 @@ namespace nadena.dev.modular_avatar.core
         [CanBeNull]
         public static string RelativePath(GameObject root, GameObject child)
         {
-            if (root == child) return "";
-
-            List<string> pathSegments = new List<string>();
-            while (child != root && child != null)
-            {
-                pathSegments.Add(child.name);
-                child = child.transform.parent?.gameObject;
-            }
-
-            if (child == null && root != null) return null;
-
-            pathSegments.Reverse();
-            return String.Join("/", pathSegments);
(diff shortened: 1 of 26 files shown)


Runtime/RuntimeUtil.cs
```diff
@@ -99,6 +89,11 @@ public static VRCAvatarDescriptor FindAvatarInParents(Transform target)
             return null;
         }
 
+        public static Transform FindAvatarTransformInParents(Transform target)
+        {
+            return ndmf.runtime.RuntimeUtil.FindAvatarInParents(target);
+        }
```
kaikoga, line 95: `RuntimeUtil.FindAvatarTransform()` is taken...

bdunderscore:
I think this can be handled with a single PR. However, I'd like to wait a few days to see if there are other critical bugs in 1.8.0 before merging this large change.

bdunderscore (approved):

PR #491, opened 2023-10-10
add HelpURL Attribute

anatawa12:
Now you can open website with clicking `?` button on the inspector

image


bdunderscore (commented):
That's a handy feature! I note that there's a mix of Japanese and english links here. I'll set up a redirect handler to detect based on your system language sometime this week, but the PR will then need to be updated to use it.

PR #504, opened 2023-10-17
Wrap VRC dependencies and VRCAvatarDescriptors

kaikoga:
Related issue: https://github.com/bdunderscore/modular-avatar/issues/478

This PR wraps everything VRCSDK specific inside `#if MA_VRCSDK3_AVATARS` and remove any `VRCAvatarDescriptor` usage outside them (some leftover of #482).

Note: Does not get rid of VRCSDK-specific runtime components (like `ModularAvatarPBBlocker` or `ModularAvatarParameters`) yet. It is more strategic than technical change in Runtime only, so I think it deserves a separate PR.

Note: VRCSDK specific MA components in prefabs would be Missing. They are supposed to be [removed on processing avatar](https://github.com/bdunderscore/modular-avatar/pull/404).


Runtime/AvatarTagComponent.cs
```diff
@@ -55,4 +58,15 @@ public virtual void ResolveReferences()
...
+#if !MA_VRCSDK3_AVATARS
+
+    /**
+     * Placeholder of VRC.SDKBase.IEditorOnly for environments without VRCSDK
+     */
+    interface IEditorOnly
+    {
+    }
+
+#endif
```
kaikoga, line 71: MEMO: should this go to ndmf? https://github.com/anatawa12/AvatarOptimizer/pull/627/files

bdunderscore:
Thanks for the PR! Sorry - I probably won't have time to review this properly until next week. I'll take a look when I find the time...

Editor/Animation/AnimatorCombiner.cs
```diff
@@ -277,13 +281,15 @@ private void AdjustBehavior(StateMachineBehaviour behavior)
         {
             switch (behavior)
             {
+#if MA_VRCSDK3_AVATARS
```
bdunderscore, line 284: Move this outside the switch for now

Editor/AnimatorMerger.cs
```diff
@@ -256,13 +260,15 @@ private void AdjustBehavior(StateMachineBehaviour behavior)
         {
             switch (behavior)
             {
+#if MA_VRCSDK3_AVATARS
```
bdunderscore, line 263: Same here (hm, these classes seem duplicated...)

Editor/Menu/VirtualMenu.cs
```diff
@@ -1,4 +1,6 @@
-﻿using System;
+﻿#if MA_VRCSDK3_AVATARS
```
bdunderscore, line 1: I wonder if it would be better to move the menu processing (or rather, the VRChat-specific stuff) to a separate asmdef and make it conditionally built...

Editor/Util.cs
```diff
@@ -29,11 +29,17 @@
...
 using VRC.SDKBase.Editor.BuildPipeline;
+#endif
+
 using Object = UnityEngine.Object;
 
 namespace nadena.dev.modular_avatar.core.editor
 {
+
+#if MA_VRCSDK3_AVATARS
     internal class CleanupTempAssets : IVRCSDKPostprocessAvatarCallback
```
bdunderscore, line 43: This is dead code and can be deleted.

bdunderscore (changes requested):

Editor/Animation/AnimatorCombiner.cs
kaikoga: Should I rather wrap entire `AdjustBehavior()` method and calls to it, or keep the `AdjustBehavior()` method empty?

Editor/Menu/VirtualMenu.cs
kaikoga: I think breaking up asmdef would deserve a separate PR before this one. I am not certain enough how to move assets into folders, though. (at least I can try... everything VRCSDK under `Runtime/VRCSDK` and `Editor/VRCSDK` I suppose?)

Editor/Util.cs
kaikoga: That would be another task before this one.

kaikoga: #515

Editor/Menu/VirtualMenu.cs
kaikoga: Couldn't try because extracting VRChat asmdef required extracting PluginDefinitions asmdef, too much to include in this PR

bdunderscore: Fair enough. We can work on that in a followup (in any case, I don't think we can move components...)

Editor/Animation/AnimatorCombiner.cs
bdunderscore: Leave the method empty for now, I think it's reasonable to keep the method call itself to avoid spreading `#if`s all over the codebase

bdunderscore (approved):

bdunderscore:
Sorry for how long the reviews took - and thanks!

PR #555, opened 2023-12-13
Added BlendTree to right click create asset menu.

rurre:
Not sure where to put it, so into the root editor folder it goes.

### DIFF
--- a/Editor/CreateBlendTree.cs
+++ b/Editor/CreateBlendTree.cs
@@ -6,14 +6,14 @@ using UnityEngine;
 
 namespace nadena.dev.modular_avatar.core.editor
 {
-    public class CreateBlendTree : Editor
+    internal static class CreateBlendTree
     {
         [MenuItem("Assets/Create/BlendTree", priority = 411)]
         static void CreateNewBlendTree()
         {
             ProjectWindowUtil.StartNameEditingIfProjectWindowExists(
                0,
-               CreateInstance<DoCreateBlendTree>(),
+               Editor.CreateInstance<DoCreateBlendTree>(),
                "New BlendTree.asset",
                EditorGUIUtility.IconContent("BlendTree Icon").image as Texture2D,
                null);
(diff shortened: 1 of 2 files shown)


Editor/CreateBlendTree.cs
```diff
@@ -0,0 +1,32 @@
+using System.IO;
+using UnityEditor;
+using UnityEditor.Animations;
+using UnityEditor.ProjectWindowCallback;
+using UnityEngine;
+
+namespace nadena.dev.modular_avatar.core.editor
+{
+    public class CreateBlendTree : Editor
```
bdunderscore, line 9: Please make this an `internal static class` as this shouldn't be an exposed API.

Editor/CreateBlendTree.cs
```diff
@@ -0,0 +1,32 @@
...
+    {
+        [MenuItem("Assets/Create/BlendTree", priority = 411)]
+        static void CreateNewBlendTree()
+        {
+            ProjectWindowUtil.StartNameEditingIfProjectWindowExists(
```
bdunderscore, line 14: Is this an undocumented API?

Editor/CreateBlendTree.cs
rurre: Done.

Editor/CreateBlendTree.cs
rurre: It's seemingly an older API. Seems to be missing from later unity version documentation. There's a page on end rename action, but not start.
https://docs.unity3d.com/ScriptReference/ProjectWindowCallback.EndNameEditAction.Action.html

But it is in use by default assets
https://github.com/Unity-Technologies/UnityCsReference/blob/496f7d6c5c0882f35bb776e96356712d54710033/Editor/Mono/ProjectBrowser/ProjectWindowUtil.cs#L347

What this ends up doing is beginning to rename a new asset, confirming the name then actually creating it, like any default asset you create.

bdunderscore:
I'll look to merging this when I have the blend tree merge component implemented.

PR #569, opened 2023-12-19
UniVRM components support

kaikoga:
Discussion: https://github.com/bdunderscore/modular-avatar/discussions/467

Adds handlings of existing UniVRM components.

This does not yet support "merging" UniVRM components, as I think merging concrete UniVRM components makes no sense and we should add components like `MA Merge VRM1 SpringBones` (which adds SpringBone entries to `VRM10Instance` ) or `MA Merge VRM0+VRM1 FirstPerson` (which adds FirstPerson settings to `VRMFirstPerson` / `VRM10Object` ) instead.

For the final commit... should this make use of ndmf Platform API once it is available?


bdunderscore:
I don't think using the platform API is required here, since this logic won't _break_ when building for VRChat. If anything, there could be a platform pass that clears UniVRM components when building for VRChat.

kaikoga:
I think I would rebase this PR because some VRM specific MA components are planned and their inspectors would need some localization.

kaikoga:
ndmf 側の様子を見て進めたいので一旦 draft にします・・・

kaikoga:
余裕があったら考える：この PR の内容をプラグインできる方法

kaikoga:
I have chosen the path to create a dedicated package for handling UniVRM stuff, hence closing this PR.

PR #575, opened 2023-12-24
fix: Replace Object dropping children (bdunderscore#547)

### LINKED ISSUE #547: Replace Object breaks Armatures when replacing bones
I have a prefab that needs to replace humanoid bones (like `UpperLeg`) to add constraints to those bones. The `Replace Object` MA components works perfectly for that.

However, this now breaks when I have another prefab (outfit) that uses `Merge Armature` to attach bones. This outfit includes new Phys Bone chains that are attached to the humanoid bone (like `UpperLeg`) being replaced by the other prefab.

Both prefabs work fine when used individually, but when used on the same avatar the following happens (my best guess):
`Merge Armature` [runs first](https://github.com/bdunderscore/modular-avatar/issues/308#issuecomment-1657120913) and correctly combines the two armatures. The `Replace Object` runs and replaces the humanoid bone. At this point, the connection between the mesh of the outfit and the moved bone under the replaced object seems to get lost. Then the remove useless object pass entirely removes the object for the new phys bone chain. This results in missing transforms on anything that was referencing the bones, and more importantly, anything that was weighted to the bones is now stuck at 0,0,0.

konsti219:
Fixes #547. See my comment there for context.

### DIFF
--- a/Editor/ReplaceObjectPass.cs
+++ b/Editor/ReplaceObjectPass.cs
@@ -83,8 +83,13 @@ namespace nadena.dev.modular_avatar.core.editor
                 replacement.transform.SetParent(original.transform.parent, true);
                 var siblingIndex = original.transform.GetSiblingIndex();
 
-                // Move children of original parent
+                // Move children of original parent (the List needs to be cloned first because it is being modified)
+                List<Transform> children = new List<Transform>();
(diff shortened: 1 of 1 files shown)


bdunderscore:
Thanks for tracking this issue down!

PR #601, opened 2023-12-30
Add MA Merge VRM1 SpringBones

kaikoga:
Discussion: https://github.com/bdunderscore/modular-avatar/discussions/467
Merge base PR first: UniVRM components support #569

image

This PR introduces `MA Merge VRM1 SpringBones` component, which adds SpringBone entries to `VRM10Instance`.
This component does nothing when the target avatar is not VRM1.

I think there is nothing specific to this component that requires localization. Perhaps there could be some warning like "This component is ignored; supports VRM1 avatar only", but many other components that targets VRCSDK only have similar issue.

I have not yet added documents. Merging this PR reveals UniVRM support to end users, so documentation would block the next release once this PR is merged.
I don't know yet where to add documents. (It would be under `docs~/`, but I worry that throwing VRM specific components into the Component reference section makes little sense to VRC users, when the entire components list is even more growing).

### DIFF
--- a/Editor/Animation/AnimationDatabase.cs
+++ b/Editor/Animation/AnimationDatabase.cs
@@ -128,6 +128,8 @@ namespace nadena.dev.modular_avatar.animation
 #if MA_VRCSDK3_AVATARS
             var avatarDescriptor = context.AvatarDescriptor;
 
+            if (!avatarDescriptor) return;
+
             foreach (var layer in avatarDescriptor.baseAnimationLayers)
             {
                 BootstrapLayer(layer);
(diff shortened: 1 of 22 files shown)


bdunderscore:
ありがとうございます！この辺見れるのは1月中旬になるので少々お待ちください

bdunderscore:
SpringBonesなどについては、クロスプラットフォーム対応に向けて要検討です・・・VRC+VRM+Resoniteで三重設定するのは避けたいので、何かのプリセット的な何かがあったほうがいい気がするんだよね・・・

kaikoga:
see #602.

PR #610, opened 2024-01-06
fix: AnimationDatabase does not register clone to ObjectRegistry

### LINKED ISSUE #609: AnimationDatabase does not register to ObjectRegistry
AnimationDatabase does not register to ObjectRegistry so some part of errors by AvatarOptimizer is broken.

With 6689c15c

anatawa12:
Fixes #609

### DIFF
--- a/Editor/Animation/AnimationDatabase.cs
+++ b/Editor/Animation/AnimationDatabase.cs
@@ -4,6 +4,7 @@ using System.Collections.Immutable;
 using System.Data.Odbc;
 using nadena.dev.modular_avatar.core.editor;
 using nadena.dev.modular_avatar.editor.ErrorReporting;
+using nadena.dev.ndmf;
 using UnityEditor;
 using UnityEditor.Animations;
 using UnityEngine;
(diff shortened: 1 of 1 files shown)


Editor/Animation/AnimationDatabase.cs
```diff
@@ -171,11 +172,33 @@ internal void RegisterState(AnimatorState state, Action<ClipHolder> processClip
...
             _clipCommitActions.Add(() => { state.motion = clipHolder.CurrentClip; });
         }
 
+        private void RegisterMotionMapping(Motion original, Motion replacement)
+        {
+            ObjectRegistry.RegisterReplacedObject(original, replacement);
+
+            if (original is BlendTree originalTree 
+                && replacement is BlendTree replacementTree 
+                && originalTree.children.Length == replacementTree.children.Length)
```
bdunderscore, line 190: Hmm. It looks like we need to duplicate the contents of the blend tree as well (we're not doing this here, which is probably an oversight...)

bdunderscore:
Thanks for the PR - Since the lack of blendtree cloning also needs to be fixed, and that would change the implementation pretty substantially, I'll just redo it at the same time.

PR #648, opened 2024-02-04
feat: allow animator-only defaults to be set using MA Parameters

bdunderscore:


bdunderscore:
TODO:
* Unit tests
* Support in 2019 UI?

PR #650, opened 2024-02-04
chore: Fix non-VRChat support

kaikoga:
Fixed some compile errors I've found when using Modular Avatar in non-VRChat project.


bdunderscore:
AnimatorCombiner is not VRC specific but there are no non-VRC use cases at this time (do other platforms expose Unity animators directly? Maybe CVR?)

PR #697, opened 2024-02-26
fix: handle AnimatorOverrideController (and null animator controllers) in Merge Animator

bdunderscore:
Closes: #692


bdunderscore:
Pending unit tests

PR #698, opened 2024-02-26
fix: untranslated warning generated when multiple parameters remap to the same name

bdunderscore:
Fixed by allowing this case and applying normal parameter type resolution rules.

Closes: #694


bdunderscore:
Pending unit tests

PR #718, opened 2024-03-02
fix: ScaleAdjuster breaks scene view selection

bdunderscore:
... fixed by reimplementing ScaleAdjuster (again!)


bdunderscore:
WIP:
* Entering play mode leaves meshes disabled
* Leaving play mode breaks bone mappings
* Loading a saved scene can have incorrect visibility settings (unity 2019?)

bdunderscore:
TODO:
* Use layer masks to control visibility?

PR #730, opened 2024-03-04
fix: improved Scale Adjuster implementation

bdunderscore:
This new implementation avoids creating any internal objects within the avatar
itself, instead creating them as hidden scene root objects. We also move all update
processing to be driven by camera callbacks as well. These changes help improve
compatibility with tools such as FaceEmo, as well as being less likely to break when
other tools manipulate the avatar's hierarchy directly.


bdunderscore:
TODO: protection against leaking renderer proxy objects

PR #731, opened 2024-03-04
chore: Fix non-VRChat support

kaikoga:
Fixed some compile errors I've found when using Modular Avatar in VRCSDK project with non-VRChat avatars, but for MA 1.9.4.

Note: Actually we don't need this PR until bdunderscore/ndmf#71 is merged, because AvatarRoot without VRCAvatarDescriptor in VRCSDK projects would not be allowed until then.


bdunderscore (approved):

bdunderscore:
I note that this isn't actually a compile error, but this is fine for a temporary hack until I can implement the NDMF platform system.

PR #828, opened 2024-04-16
feat: add support for A-pose/T-pose conversion in Setup Outfit

bdunderscore:


bdunderscore:
Before:
image

After: 
image

PR #881, opened 2024-06-12
Fix links in Japanese docs

chigirits:
Links between pages within the Japanese document were linked to the English version. This PR fixes that.
I just modified the Markdown and have not rebuilt it to check, but if there is no problem, please merge this PR.


bdunderscore:
This is a pain to test; since the build-time link check worked I'll merge it, but if it breaks things on the dev docs site I'll need to roll it back.

Issue #858, opened 2024-05-26
VRC Animator Play Audio Source Rename


BigAtomikku:
Currently Modular Avatar does not automatically rename the "source" field inside VRC Animator Play Audio when using Relative paths.

bdunderscore:
Duplicate of https://github.com/bdunderscore/modular-avatar/issues/783

Issue #783, opened 2024-03-16
Better support for Play Audio State Behavior


bdunderscore:
We'd like to support object path remapping, and maybe an automatic audio source pool system...

PR #899, opened 2024-06-23
Implement VRCAnimatorPlayAudio handling logic

JLChnToZ:
This is my attempt to implement handling source path inside `VRCAnimatorPlayAudio` for Modular Avatar. (#783, #858)
It is a quite dirty approach and don't know if the code quality is qualifiable to merge into upstream so I file a PR here anyways.

Before Animator Merging:
image

After Animator Merging:
image

### DIFF
--- a/Editor/Animation/AnimationDatabase.cs
+++ b/Editor/Animation/AnimationDatabase.cs
@@ -80,37 +80,12 @@ namespace nadena.dev.modular_avatar.animation
             }
         }
 
-#if MA_VRCSDK3_AVATARS_3_5_2_OR_NEWER
-        internal class PlayAudioHolder
-        {
-            private VRCAnimatorPlayAudio _currentPlayAudio;
-
-            internal VRCAnimatorPlayAudio CurrentPlayAudio
-            {
-                get
-                {
-                    return _currentPlayAudio;
-                }
-                set
-                {
-                    _currentPlayAudio = value;
-                }
-            }
-
-            internal PlayAudioHolder(VRCAnimatorPlayAudio audio)
-            {
-                CurrentPlayAudio = audio;
-            }
-        }
-#endif
-
-
         private BuildContext _context;
 
         private List<Action> _clipCommitActions = new List<Action>();
         private List<ClipHolder> _clips = new List<ClipHolder>();
 #if MA_VRCSDK3_AVATARS_3_5_2_OR_NEWER
-        private List<PlayAudioHolder> _playAudios = new List<PlayAudioHolder>();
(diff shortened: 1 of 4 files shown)


Editor/Animation/AnimationDatabase.cs
```diff
@@ -80,10 +80,38 @@ public void SetCurrentNoInvalidate(Motion newMotion)
...
+                {
+                    _currentPlayAudio = value;
+                }
+            }
+
+            internal PlayAudioHolder(VRCAnimatorPlayAudio audio)
+            {
+                CurrentPlayAudio = audio;
+            }
+        }
```
bdunderscore, line 104: We probably don't need a holder class for this and can just reference the VRCAnimatorPlayAudio assets directly. The reason we have the holder for clips is because we need to swap out clips for different clips in processing (in part due to the lack of an API to delete editor curve bindings), but with VRCAnimatorPlayAudio the state behaviors are already cloned at this point and we can just update the paths in place.

Editor/Animation/PathMappings.cs
```diff
@@ -274,6 +278,16 @@ private string MapPath(EditorCurveBinding binding)
             return newClip;
         }
 
+#if MA_VRCSDK3_AVATARS_3_5_2_OR_NEWER
+        private VRCAnimatorPlayAudio ApplyMappingsToPlayAudio(VRCAnimatorPlayAudio audio)
+        {
+            Debug.Log($"SourcePath: {audio.SourcePath}");
```
bdunderscore, line 284: Please remove the Debug.Log statements before merging (I know, I'm bad at it too sometimes... 😅 )

Editor/nadena.dev.modular-avatar.core.editor.asmdef
```diff
@@ -43,6 +43,11 @@
             "name": "com.vrchat.avatars",
             "expression": "",
             "define": "MA_VRCSDK3_AVATARS"
+        },
+        {
+            "name": "com.vrchat.avatars",
+            "expression": "3.5.2",
+            "define": "MA_VRCSDK3_AVATARS_3_5_2_OR_NEWER"
```
bdunderscore, line 50: OK for now but I'll probably make this a minimum requirement before 1.10.0 release.

bdunderscore (changes requested):
Thanks for looking into this! Some small comments, but otherwise looks good. FYI, though - the preview system and Shape Changer stuff has a bit more work to go so release probably won't be for another few weeks.

bdunderscore (approved):

bdunderscore:
Thanks for the contribution! For future reference, please create a branch on your repository when you submit a PR, it makes it easier to make small edits before merging (I'll be following up with some unit tests and etc)

PR #1008, opened 2024-08-14
feat: menu name may use string field

hai-vr:
This is a proposal to include a new "label" field in Modular Avatar Menu Item.

Although not specified, menu items may contain line breaks and [rich text](https://docs.unity3d.com/Packages/com.unity.ugui@1.0/manual/StyledText.html), which will be displayed in the radial menu.

It is currently really awkward to write rich text within the object name. In addition to being long, rich text tags will contain a forward slash `/` and other symbols.

Exposing that field may enable other inventory automation tools to more precisely control the generation of the expression menu items.

This PR attempts to add the following:
- Adds a new serializable field "label" in the Modular Avatar Menu Item.
- When this "label" field is not an empty string, then Modular Avatar Menu Item would use that instead of the GameObject name.
- In the UI, add some means to enable that field for editing, add a multi-line field for that label, and a rich text preview area.

The UI method is currently unsatisfactory, I do not know what would be a good fit. Most users will not use rich text, so the current proposal hides the existence of that new field to the user unless they click a "Rich text" checkbox, which temporarily enables the visibility of the field even when that field is an empty string.

image

VRChat_LCLEUVAbML


bdunderscore:
Perhaps a better way to handle this would be to have the label default to the gameobject name when the field is empty?
To help make this behavior clearer, we could display the gameobject name in greyed-out text over the label field when not set.

bdunderscore:
We could also only show the preview if the text contains `<`...

Editor/Inspector/Menu/MenuItemGUI.cs
```diff
@@ -215,11 +220,46 @@ public void DoGUI()
...
-            EditorGUI.BeginChangeCheck();
-            EditorGUILayout.PropertyField(_name, G("menuitem.prop.name"));
-            if (EditorGUI.EndChangeCheck())
+
+            EditorGUILayout.BeginHorizontal();
+            var needsRichLabel = (!string.IsNullOrEmpty(_prop_label.stringValue) || _isTryingRichLabel);
+            if (!needsRichLabel)
+            {
+                EditorGUI.BeginChangeCheck();
+                EditorGUILayout.PropertyField(_name, G("menuitem.prop.name"));
```
bdunderscore, line 229: If we're linked, we should just show the name property (which can link directly to the GameObject if this is a `ModularAvatarMenuItem`). If we're not linked, we should show the label field always (and conditionally show the rich text preview).

Editor/Inspector/Menu/MenuItemGUI.cs
```diff
@@ -215,11 +220,46 @@ public void DoGUI()
...
+                if (EditorGUI.EndChangeCheck())
+                {
+                    _name.serializedObject.ApplyModifiedProperties();
+                }
+            }
+            else
+            {
+                EditorGUILayout.PropertyField(_prop_label, G("menuitem.prop.name"));
+            }
+            var linkIcon = EditorGUIUtility.IconContent(needsRichLabel ? "UnLinked" : "Linked").image;
```
bdunderscore, line 239: This UI class can be used to render VRCExpressionMenu controls out of expression menu assets as well, in this case the link icon doesn't make sense, as there's no GameObject to link to. See the `MenuItemCoreGUI(GameObject parameterReference, SerializedProperty _control, Action redraw)` constructor for details.

Editor/Inspector/Menu/MenuItemGUI.cs
```diff
@@ -215,11 +220,46 @@ public void DoGUI()
...
+            var guiIcon = new GUIContent(linkIcon, S(needsRichLabel ? "menuitem.rich_text.toggle_off.tooltip" : "menuitem.rich_text.toggle_on.tooltip"));
```
bdunderscore, line 240: We can just enable this when you type a `<`

bdunderscore: Also, this option doesn't directly control rich text, but rather whether we're linking our name to the gameobject, so the translation key (and text) should probably reflect that.

Editor/Inspector/Menu/MenuItemGUI.cs
```diff
@@ -215,11 +220,46 @@ public void DoGUI()
             EditorGUILayout.BeginHorizontal();
 
             EditorGUILayout.BeginVertical();
-            EditorGUI.BeginChangeCheck();
-            EditorGUILayout.PropertyField(_name, G("menuitem.prop.name"));
-            if (EditorGUI.EndChangeCheck())
+
+            EditorGUILayout.BeginHorizontal();
+            var needsRichLabel = (!string.IsNullOrEmpty(_prop_label.stringValue) || _isTryingRichLabel);
```
bdunderscore, line 225: Would it be simpler to just do `_isTryingRichLabel |= !string.IsNullOrEmpty(_prop_label.stringValue)`?

Editor/Inspector/Menu/MenuItemGUI.cs
```diff
@@ -215,11 +220,46 @@ public void DoGUI()
             EditorGUILayout.BeginHorizontal();
 
             EditorGUILayout.BeginVertical();
-            EditorGUI.BeginChangeCheck();
-            EditorGUILayout.PropertyField(_name, G("menuitem.prop.name"));
-            if (EditorGUI.EndChangeCheck())
+
```
bdunderscore, line 223: One note: Please make sure this doesn't break in some terrible way when multiple items are selected...

bdunderscore (changes requested):

hai-vr:
For the next commit, I had to rebase to 1.10.0-rc.4, because it was not possible to test Sub Menu of source Expressions Menu Asset on the base I started on

Editor/Inspector/Menu/MenuItemGUI.cs
bdunderscore: It's maybe better to multi-edit the label field in that case. Seems like less surprising behavior than editing the name of all the GameObjects at once?

hai-vr: Should be current behaviour as of the latest commit

Spokeek:
I like that. Great idea !

bdunderscore (commented):
Thanks for the patch! I'll include this in the next minor update (1.11) - probably in the next week or so?

Note that there's a minor bit of UI weirdness that might be good to fix while waiting - see my comment below.

bdunderscore (approved):

PR #1074, opened 2024-09-01
feat: update AvatarObjectReference paths when target object is moved in scene

bdunderscore:
Closes: #1037


bdunderscore:
Draft: pending NDMF changes

PR #1232, opened 2024-10-01
Make compatible with Unity 6 projects

hai-vr:
This PR intends to make some Modular Avatar 1.10 components usable within the [Basis](https://github.com/dooly123/Basis) framework, which is currently using Unity 6 (6000.0.20f1).


bdunderscore:
I'll take a look at this probably over the weekend - I'm prioritizing postlaunch bug fixes for now

Editor/ReactiveObjects/ParameterAssignerPass.cs
```diff
@@ -1,4 +1,5 @@
-﻿using System.Collections.Generic;
+﻿#if MA_VRCSDK3_AVATARS
```
bdunderscore, line 1: It might be cleaner to break out the ReactiveObjects directory into a different assembly and conditionally compile it...

bdunderscore (changes requested):

bdunderscore (approved):

Issue #1235, opened 2024-10-01
Pass Harmony ID to UnpatchAll


bdunderscore:
https://discord.com/channels/1001423276553285653/1001424777401077830/1290634990832320676

PR #1241, opened 2024-10-02
chore: set harmony ID on UnpatchAll call

bdunderscore:


bdunderscore:
Partially closes #1235  (we still need to do NDMF as well)

PR #1313, opened 2024-10-21
Fixes error when merging same parameter with different type in RC menu item

### LINKED ISSUE #1299: Merging an animation using the same parameter name as an RC menu item, but non-float type, results in an error

JLChnToZ:
This PR attempts to fixes #1299 

The issue causing type mismatch is it doesn't do type checking when generating new states, and re-using existing AnimatorCombiner class might be just good enough here.

(P.S. This is a quick fix as this issue also affects Hatch Menu, one of my gimmick with an installer that uses Modular Avatar)

### DIFF
--- a/Editor/ReactiveObjects/AnimationGeneration/ReactiveObjectPass.cs
+++ b/Editor/ReactiveObjects/AnimationGeneration/ReactiveObjectPass.cs
@@ -26,10 +26,7 @@ namespace nadena.dev.modular_avatar.core.editor
         
         private AnimationClip _initialStateClip;
         private bool _writeDefaults;
-
-        private AnimatorCombiner _combineSession;
-        private AnimatorController _animController;
-
+        
         public ReactiveObjectPass(ndmf.BuildContext context)
         {
             this.context = context;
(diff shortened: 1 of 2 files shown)


Editor/ReactiveObjects/AnimationGeneration/ReactiveObjectPass.cs
```diff
@@ -26,34 +26,48 @@ internal partial class ReactiveObjectPass
...
         {
+            var fxController = FindFxController();
+
             // Having a WD OFF layer after WD ON layers can break WD. We match the behavior of the existing states,
             // and if mixed, use WD ON to maximize compatibility.
-            _writeDefaults = MergeAnimatorProcessor.ProbeWriteDefaults(FindFxController().animatorController as AnimatorController) ?? true;
-            
+            _writeDefaults = MergeAnimatorProcessor.ProbeWriteDefaults(fxController.animatorController as AnimatorController) ?? true;
+
+            _combineSession = new AnimatorCombiner(context, fxController.animatorController != null ? fxController.animatorController.name : "FX (Reactive Objects)");
```
bdunderscore, line 46: Recreating the AnimatorCombiner while AnimationServicesContext is active will break the commit actions held in the AnimationDatabase, which holds a series of delegates which write the modified or replaced Motions back to the states they came from: https://github.com/bdunderscore/modular-avatar/blob/main/Editor/Animation/AnimationDatabase.cs#L100

Probably a better short-term approach would be to detect which parameters need this, and force them to be floats during Merge Animator processing. Longer-term I'm working on a redesign of the AnimationServicesContext and AnimatorCombiner to play nicer together.

bdunderscore (changes requested):

Editor/ReactiveObjects/AnimationGeneration/ReactiveObjectPass.cs
bdunderscore: It might be enough to add logic to `ParameterAssignerPass` to generate a Merge Animator component, which merges an animator that declares all relevant parameters (as floats) but has zero layers...

Editor/ReactiveObjects/ParameterAssignerPass.cs
```diff
@@ -186,6 +187,19 @@ protected override void Execute(ndmf.BuildContext context)
...
+            {
+                // This make sures the parameters are correctly merged into the FX layer.
+                context.AvatarRootObject.AddComponent<ModularAvatarMergeAnimator>().animator = new AnimatorController
+                {
+                    parameters = _mamiByParam.Keys.Select(name => new AnimatorControllerParameter
+                    {
+                        name = name,
+                        type = AnimatorControllerParameterType.Float,
+                    }).ToArray(),
+                };
```
bdunderscore, line 201: As a best practice, please be specific about relevant configuration (e.g., target animator layer = FX, remove attached animator = false)

bdunderscore:
Thanks for the fix!

bdunderscore (approved):

PR #1326, opened 2024-10-30
Use VRCParentConstraint instead of constraint hack for world fixed objects when available

JLChnToZ:
As title indicates, this patches `WorldFixedObjectProcessor` to use `VRCParentConstraint` instead of old-fashioned Unity constraint hack for world fixed objects, only when such components are available and we're working on VRChat avatars.


Editor/WorldFixedObjectProcessor.cs
```diff
@@ -86,16 +86,38 @@ private Transform CreateProxy()
...
+                constraint.AffectsPositionX = true;
+                constraint.AffectsPositionY = true;
+                constraint.AffectsPositionZ = true;
+                constraint.AffectsRotationX = true;
+                constraint.AffectsRotationY = true;
+                constraint.AffectsRotationZ = true;
+                constraint.FreezeToWorld = true;
+            }
+            else
+#endif
```
bdunderscore, line 108: I'd prefer here to extract this logic to a method and swap out the method as a whole, rather than having an else block spanning an `#endif` like this...

JLChnToZ: IDK if the committed change is OK, it is hard to extract as a method because the case is "when VRCSDK with constraints is imported" (only can be filtered with `#if` preprocessor) and "we are working on a VRChat avatar" (only can be checked with `if (...) ... else` block)

bdunderscore: I'm thinking something like:

```
if (!TryGenerateVRChatConstraint(obj)) GenerateGenericConstraint(obj);

// ...
#if MA_VRCSDK3_AVATARS_3_7_0_OR_NEWER
bool TryGenerateVRChatConstraint(GameObject obj) { ... }
#else
bool TryGenerateVRChatConstraint(GameObject obj) { return false; }
#endif
```

JLChnToZ: OK.

bdunderscore:
It looks like the unit tests are broken now...

JLChnToZ:
Yes it requires changes on unit test, it is coded to check if `ParentConstraint` exists...

https://github.com/JLChnToZ/modular-avatar/blob/7d64d6b7a9180142db488882068394bebea293f6/UnitTests~/WorldFixedObjectTest/WorldFixedObjectTest.cs#L55

I'm not very familiar on modifying unit tests, especially with such conditional flow.

bdunderscore (approved):

PR #1363, opened 2024-11-20
Revert "Use VRCParentConstraint instead of constraint hack for world fixed objects when available (#1326)"

bdunderscore:
This reverts commit a984cf86735170d075f8224dd7c83fa760e54eed. The prior
behavior was to lock world fixed objects at their offset from the
origin; however with this change we ended up locking them at a location
relative to the avatar spawn location, breaking some gimmicks.

I tried experimenting a bit with VRCConstraint to try to replicate this
behavior, but it seems a bit non-trivial, so we'll revert this as a
hotfix for now.

### DIFF
--- a/Editor/WorldFixedObjectProcessor.cs
+++ b/Editor/WorldFixedObjectProcessor.cs
@@ -86,15 +86,6 @@ namespace nadena.dev.modular_avatar.core.editor
             obj.transform.localRotation = Quaternion.identity;
             obj.transform.localScale = Vector3.one;
 
-            if (!TryCreateVRCConstraint(avatarRoot, obj)) CreateConstraint(obj, fixedGameObject);
-
-            _proxy = obj.transform;
-
-            return obj.transform;
-        }
-
-        private void CreateConstraint(GameObject obj, GameObject fixedGameObject)
-        {
             var constraint = obj.AddComponent<ParentConstraint>();
             constraint.AddSource(new ConstraintSource()
             {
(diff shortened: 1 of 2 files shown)


bdunderscore:
FYI @JLChnToZ - this change broke some gimmicks, so I'm reverting it for now.

JLChnToZ:
I would like to know the details on when it breaks. Guessing is it breaks when mixing use of other Unity constraints without migrating?

anatawa12:
AFAIK, world origin prefab trick will work with VRC Constraints so reimplementing without FreezeToWorld would be the solution to this breakage.

JLChnToZ:
> AFAIK, world origin prefab trick will work as before with VRC Constraints so reimplementing without FreezeToWorld would be the solution to this breakage.

Do you mean just use the `VRCParentConstraint` with the prefab referenced in it?

anatawa12:
Sorry for late response.

> Do you mean just use the VRCParentConstraint with the prefab referenced in it?

Yes.

JLChnToZ:
If so, I can make another patch with this alternative logic.

JLChnToZ:
Anyways, I need to know how to reproduce the not working cases for further patching.

anatawa12:
I think just placing something under MA World Fixed without constraints to the Avatar hierarchy would reproduce the problem. (not tested)

I think old (Parent Constraint to the prefab) way would place the object at the world origin and new (FreezeToWorld) method would place the object at the spawn position.

Narazaka:
I just tried it and it seemed like the world origin was fixed with the same prefab hack as the old one.
It is possible that the situation has changed over time, so you may want to try again.
(Not from the editor UI, but generated by the following script.)

さっき試したら旧来と同様？のprefabハックでそれっぽくワールド原点固定されてる様な気がしました。
時を経て状況が変わっていることもあり得るので再度試してみてもいいかも？
（エディタからではなく下記のようなスクリプトで生成したやつですが）

```Csharp
            var c = root.AddComponent<VRCParentConstraint>();
            c.AffectsPositionX = true;
            c.AffectsPositionY = true;
            c.AffectsPositionZ = true;
            c.AffectsRotationX = true;
            c.AffectsRotationY = true;
            c.AffectsRotationZ = true;
            c.Locked = true;
            // c.FreezeToWorld = true; // sometimes breaks?
            c.IsActive = true;
            c.GlobalWeight = 1;
            var pf = AssetDatabase.LoadAssetAtPath<GameObject>("Assets/Origin.prefab");
            c.Sources.Add(new VRC.Dynamics.VRCConstraintSource(pf.transform, 1));
```

PR #1429, opened 2025-02-01
feat: support merging humanoid bones with PBs in limited cases

bdunderscore:
This change adds support for merging humanoid bones that are a target of PhysBones, provided that all humanoid children are excluded from that PhysBone (either with a direct ignores field, or using PB Blocker).

Note: Because this is a significant expansion of support, this will need to wait for a minor release to maintain semver semantics.

Closes: #1406


bdunderscore:
https://discord.com/channels/1201675918956056627/1346443440526983219/1347753194012413992

PR #1483, opened 2025-03-11
fix: Merge Animator should not adjust WD in single state or additive layers

bdunderscore:


bdunderscore:
Pending https://github.com/bdunderscore/ndmf/pull/549

yelandiealma:
I want to know why.
Btw, I think this is an important change that should be documented. Otherwise, it might cause confusion.

bdunderscore:
Setting direct blend trees or additive layer states to WD OFF can break them, and in single state layers the difference isn't (as far as I know) meaningful.

ColorlessColor:
All my avatar is wd on, but some prefab provide a single-layer single-state animator controller with wd off, which broken my wd setting. Now I can't turn my tail back on, and I have to re-upload all my avatars.

Issue #1396, opened 2024-12-18
Facial expressions not working in MMD world


fz6m:
When using `Object Toggle`, `Menu Item`, and `Menu Installer`simultaneously in the MMD world, the avatar's facial expressions fail to work as expected. ( All `Write Defaults` are forced on )

This issue does not occur in version `v1.10.0` , but it occurs in version `v1.10.1` or later.

<img src='https://github.com/user-attachments/assets/f231e40c-6fd3-41b8-8e7a-5006a2e88d7b' width='30%' />

PR #1489, opened 2025-03-12
feat: ensure that correct layers are toggled off in MMD worlds, even after messing with layer order

bdunderscore:
We make the assumption that the MMD world will _specifically_
be disabling layers 1 and 2.

Closes: #1309, #1396


bdunderscore:
https://github.com/bdunderscore/ndmf/pull/550 <-- dependency

Issue #1541, opened 2025-04-05
All my avatar is wd on, but some prefab provide a single-layer single-state animator controller with wd off, which broken my wd setting. Now I can't turn my tail back on, and I have to re-upload all my avatars.


bdunderscore:
All my avatar is wd on, but some prefab provide a single-layer single-state animator controller with wd off, which broken my wd setting. Now I can't turn my tail back on, and I have to re-upload all my avatars.

Update: Also broke the expressions in the MMD world.

_Originally posted by @ColorlessColor in https://github.com/bdunderscore/modular-avatar/issues/1483#issuecomment-2780160315_

PR #1542, opened 2025-04-05
fix: incorrect WD matching handling for single-animation-clip layers

bdunderscore:
Evidently, unity has special handling specifically for single-state layers with a blend tree, but _not_ for single-state layers with an animation clip.

Closes: #1541


ColorlessColor:
What happens if someone set the blendtree to wd off and expects the `MA Merge Animator` to set it to wd on?
I think this breaks a lot of assumptions about prefabs.

ColorlessColor:
For example, https://childwitch.booth.pm/items/6113734

Bad ABT FX Blendtree
Bad ABT FX Blendtree 2

ColorlessColor:
In general, 

- If the avatar is wd on, all states should be set to wd on;
- If the avatar is wd off, perhaps the asset creator should be ensuring the blendtree state is  wd on.

PR #1570, opened 2025-04-29
fix: workaround unity bug with nested armatures

bdunderscore:


bdunderscore:
This doesn't fully work around the issue unfortunately.

PR #1580, opened 2025-05-13
docs: documentation updates for resonite support

bdunderscore:
Also marks ALCOM as the recommended installation method


bdunderscore:
Pending NDMF release

PR #1660, opened 2025-07-26
fix: NaNimation initial states are not applied when animations are inactive

bdunderscore:
Closes: #1657


bdunderscore:
Will merge after #1651 to minimize conflicts

PR #1677, opened 2025-08-11
feat(i18n): adding simplified Chinese translation for documentation

Kagenui-yozuru:


bdunderscore:
Hmm. While I appreciate the translation, it's a bit difficult for me to review this 😅 
How are you thinking of keeping this up to date and tracking changes to the underlying documentation on an ongoing basis?

yozurukt:
If this PR is merged, I will create a workflow in my forked repository to periodically fetch updates. It will monitor for changes in the `docs/~ `directory and send me a notification. I will then submit new PR

bdunderscore:
I see. Up until now, I've been ensuring that the EN and JP documentation stay up to date by... just always updating both. This obviously doesn't scale beyond the languages I know, and makes it hard for anyone else to contribute, so I've been wanting a better system for keeping things in sync.

As-is, I don't think I can merge this until we have a good plan for how to keep it in sync without relying on an external repository, or a single external contributor; similarly to how I brought in crowdin to handle strings translations, we'll need some kind of mechanism to 1) allow for tracking of out-of-date translations and 2) review of translations, that doesn't rely on a single specific person to keep running.

Do you have any suggestions for how to best do this? I believe there's some kind of support for markdown translation in docusaurus but I haven't looked into it in much detail.

yozurukt:
After reviewing the Docusaurus documentation, I noticed there isn’t a [built-in way](https://docusaurus.io/docs/next/i18n/git#maintenance) to keep translated content in sync, since Markdown content is unstructured and relies on manual updates.
My proposed solution involves using GitHub Actions:

- Monitors changes to the English docs

- Flags outdated translations by either:

  - Opening an issue to notify maintainers, or

  - Adding a visible note in the translated file (e.g., "This translation is outdated. Refer to [English version] for updates.")

If this approach sounds reasonable, I’d be happy to draft the workflow file.

Regardless, I’ll continue maintaining the Chinese translation. If, for any reason. If maintaining it becomes unsustainable, we can revisit whether to keep or remove it.

bdunderscore:
Yes, I think that workflow would be helpful - with that in place I think we can merge it.

bdunderscore:
Closing as we don't have a workflow to help keep things in sync in place still.

Issue #1401, opened 2024-12-21
Reactive Components are overwritten by pre-existing animators when they are always-on


bdunderscore:
https://discord.com/channels/1201675918956056627/1319646329785548820/1319646329785548820

PR #1732, opened 2025-09-01
fix: static reactive components would have lower effective priority than the FX animators

bdunderscore:
Closes: #1727


nekobako:
https://github.com/bdunderscore/modular-avatar/issues/1401
過去にも同様の報告があった模様…？

PR #1779, opened 2025-10-01
Fix compilation in Unity 6

hai-vr:
This PR makes parts of Modular Avatar usable in a Basis Framework Unity 6.2 project.

Notes: I have completely covered ReactiveObjectPass.cs with a conditional flag. First of all, the pass requires an AvatarDescriptor, so it would not have run in the first place. The code within ReactiveObjectPass does imply that there ought to be some support for non-VRChat apps, but many of the classes that ReactiveObjectPass depends on are not available, see ReactiveObjectAnalyzer, which is already entirely covered with a conditional flag.

That's why I think it's fair to mute the entire class as well as the "Reactive Components" step in the sequence.

### DIFF
--- a/Editor/Inspector/MaterialSwap/MatSwapEditor.cs
+++ b/Editor/Inspector/MaterialSwap/MatSwapEditor.cs
@@ -99,7 +99,11 @@ namespace nadena.dev.modular_avatar.core.editor
                         fromProperty.serializedObject.ApplyModifiedProperties();
                     });
                 }
+#if UNITY_6000_0_OR_NEWER
+                menu.DropDown(_fromField.worldBound, _fromField, false);
+#else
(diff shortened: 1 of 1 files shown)


Editor/ReactiveObjects/AnimationGeneration/ReactiveObjectPass.cs
```diff
@@ -1,4 +1,5 @@
-﻿#region
```
bdunderscore: This is intended to function (with limitations - static states only) on a standalone project.

Toys0125:
Found a bug when building android that Unity 6 just removes System.Collections.Immutable;

bdunderscore:
I _think_ this was fixed in a few other PRs. Please feel free to open a new PR with any additional changes needed.

PR #1785, opened 2025-10-08
Add discord icon to ReadMe

JuzoVR:
I struggled finding the discord for this server so I wanted to add that to the documentation.

Added discord icon using https://shields.io/badges/discord and grabbed the invite link from bd_ server to use it. Also grabbed the server id for the MA server from the server.


JuzoVR:
Just want to emphasize that i dont need this to be merged and you are more than welcome to do your own version. I do not think I deserve a contributor badge or anything of the like for just a readme edit. I just had trouble finding the discord (yes its on the home page of the documentation, but it would be nice here)

Issue #688, opened 2024-02-21
Feature Request: Component for reassigning VRC humanoid Colliders


ZenithVal:
The hand and finger colliders on your avatar that can interact with other players' physbones can be re-assigned if you open the descriptor component in inspector debug. One popular gimmick is assigning the ring finger bone to the foot, so they can interact with physbones using it. 

D4rk's optimizer has a simple toggle for assigning the ring finger to the foot collider. [Link]( https://github.com/d4rkc0d3r/d4rkAvatarOptimizer/blob/6f8a76985b4783f06e6cf21102a0dfe6bffb11bc/Editor/d4rkAvatarOptimizer.cs#L4523)

The only issue is that if you exit debug mode and the collier section is still open, the SDK will revert these bones to their humanoid default. D4rk's optimizer gets around this by applying it with a script without ever opening the descriptor collider section. (Non destructively too) 

Other example use cases:
 - Colliders for props so they affect Physbones.
 - Assigning the collider to a constraint so you can move it anywhere on a whim.
   - I toggle my hand between my mouth and hand ~ allowing me to grab physbones with it (Bite!)

Basically, a component for full custom assignments of the colliders. Adjusting their size with a preview may also be appreciated since that's not possible in debug mode.

PR #1786, opened 2025-10-11
Global Collider

ZenithVal:
Resolves [#688](https://github.com/bdunderscore/modular-avatar/issues/688) (Closing my own issue ~ joy!)

For setting up colliders which can globally interact with other Physbones or modularizing base colliders on VRChat avatars. For VRChat, this allows reassigning & reshaping of the built in descriptor colliders. It handles the finger colliders so they can be placed like any other collider. Open for integration with other platforms.


Runtime/ModularAvatarGlobalCollider.cs
```diff
@@ -0,0 +1,58 @@
...
+
+	[AddComponentMenu("Modular Avatar/MA Global Collider")]
+	[HelpURL("https://modular-avatar.nadena.dev/docs/reference/global-collider?lang=auto")]
+	public class ModularAvatarGlobalCollider : AvatarTagComponent
+	{
+		//Toggle for manual remapping, if false auto remapping.
+		public bool manualRemap = false;
+
+		//descriptor collider to replace (Only used in manual mode)
+		public GlobalCollider sourceCollider;
```
bdunderscore: source and target can be a bit ambiguous. It's better to use something like `colliderToHijack` to make it clear which is which

Runtime/ModularAvatarGlobalCollider.cs
```diff
@@ -0,0 +1,58 @@
...
+
+		//May be better to do this with the method bone proxy uses
+		public AvatarObjectReference rootTransform = new AvatarObjectReference();
+		public GameObject remapTargetObject => rootTransform.Get(this);
```
bdunderscore: This is fine, if the goal is to be analogous with physbone references. I would suggest something more like this though:

```
internal AvatarObjectReference m_rootTransform = new();
public Transform RootTransform {
  get => m_rootTransform?.Get(this)?.transform ?? transform,
  set {
    if (m_rootTransform == null) m_rootTransform = new();
    m_rootTransform.set(value?.gameObject);
  }
}
```

Runtime/ModularAvatarGlobalCollider.cs
```diff
@@ -0,0 +1,58 @@
...
+			if (RuntimeUtil.IsResetFromInspector())
+			{
+				InitSettings();
+			}
+		}
+
+		internal void InitSettings()
+		{
+			sourceCollider = GlobalCollider.FingerRingLeft;
+		}
```
bdunderscore: I would prefer to leave this unset (and require the user to be opinionated in manual mode)

Runtime/ModularAvatarGlobalCollider.cs
```diff
@@ -0,0 +1,58 @@
...
+		FootLeft, FootRight,
+		None,
+	}
+
+	[AddComponentMenu("Modular Avatar/MA Global Collider")]
+	[HelpURL("https://modular-avatar.nadena.dev/docs/reference/global-collider?lang=auto")]
+	public class ModularAvatarGlobalCollider : AvatarTagComponent
+	{
+		//Toggle for manual remapping, if false auto remapping.
+		public bool manualRemap = false;
```
bdunderscore: For new components, I prefer to use a combination of an internal field (naming scheme `m_fieldName`), and a public property with getter and setter. This is because replacing a field with a property is not an API-compatible change; wrapping with a property from the start preserves the option of adding more logic here in the future.

Runtime/ModularAvatarGlobalCollider.cs
```diff
@@ -0,0 +1,58 @@
...
+
+		//descriptor collider to replace (Only used in manual mode)
+		public GlobalCollider sourceCollider;
+
+		//May be better to do this with the method bone proxy uses
+		public AvatarObjectReference rootTransform = new AvatarObjectReference();
+		public GameObject remapTargetObject => rootTransform.Get(this);
+
+		public bool copyOriginalShape = false;
```
bdunderscore: Is this only valid in manual mode?

Runtime/ModularAvatarGlobalCollider.cs
```diff
@@ -0,0 +1,58 @@
...
+		public bool visualizeGizmo = true;
+		public float radius = 0.05f;
+		public float height = 0.2f;
+		public Vector3 position = Vector3.zero;
+		public Quaternion rotation = Quaternion.identity;
```
bdunderscore: Somewhere we should document the axis the capsule's height corresponds to (is it Y or Z?)

Editor/GlobalColliderPass.cs
```diff
@@ -0,0 +1,260 @@
...
+namespace nadena.dev.modular_avatar.core.editor
+{
+	internal class GlobalColliderPass : Pass<GlobalColliderPass>
+	{
+		protected override void Execute(ndmf.BuildContext ctx)
+		{
+			var remapColliders = ctx.AvatarRootTransform.GetComponentsInChildren<ModularAvatarGlobalCollider>(true);
+			if (remapColliders.Length == 0) return;
+
+			Array.Sort(remapColliders, (a, b) =>
```
bdunderscore: Array.Sort is not a stable sort (so the results will vary from one build to the next). Use `OrderBy` instead: https://learn.microsoft.com/en-us/dotnet/api/system.linq.enumerable.orderby?view=net-9.0&redirectedfrom=MSDN#System_Linq_Enumerable_OrderBy__2_System_Collections_Generic_IEnumerable___0__System_Func___0___1__

This also is a bit cleaner as you can use a key mapping function instead of a manual comparator.

Editor/GlobalColliderPass.cs
```diff
@@ -0,0 +1,260 @@
...
+				if (a.manualRemap && !b.manualRemap) return -1;
+				if (!a.manualRemap && b.manualRemap) return 1;
+				return 0;
+			});
+
+#if MA_VRCSDK3_AVATARS
+			var indexFingerWarns = new List<GameObject>();
+			var skippedColliders = new List<GameObject>();
+
+			var usedColliders = new List<GlobalCollider>();
```
bdunderscore: HashSet?

Editor/GlobalColliderPass.cs
```diff
@@ -0,0 +1,260 @@
...
+				};
+
+				if (my.manualRemap)
+				{
+					//This ended up being overly instrusive since it's purposefully used in some of my workflows. I don't think it's needed.
+					/*if (usedColliders.Contains(my.sourceCollider))
+					{
+						//Collider was already used, so it will be overriten by this remap.
+						//BuildReport.Log(ErrorSeverity.Information, "validation.global_collider.manual_collider_overwrite", my.gameObject);
+					}*/
```
bdunderscore: This should probably be a warning. It would be hard to debug otherwise, after all (and it should cite both components, not just one)

Editor/GlobalColliderPass.cs
```diff
@@ -0,0 +1,260 @@
...
+							break;
+					}
+					newColliderConfig.radius = sourceColliderConfig.radius;
+					newColliderConfig.height = sourceColliderConfig.height;
+					newColliderConfig.position = sourceColliderConfig.position;
+					newColliderConfig.rotation = sourceColliderConfig.rotation;
+				}
+
+				//Apply new collider config to descriptor
+				switch (targetCollider)
```
bdunderscore: Can we combine the two switch statements using e.g. a `ApplyCollider(ref desc.collider_footR, false)`

Editor/GlobalColliderPass.cs
```diff
@@ -0,0 +1,260 @@
...
+			}
+
+			if (indexFingerWarns.Count > 0)
+			{
+				BuildReport.Log(ErrorSeverity.Information, "validation.global_collider.using_index_fingers_vrc", indexFingerWarns);
+			}
+			if (skippedColliders.Count > 0)
+			{
+				BuildReport.Log(ErrorSeverity.NonFatal, "error.global_collider.no_global_colliders_available_vrc", skippedColliders);
+			}
```
bdunderscore: I do wonder if we should just list all global colliders - when we hit these errors we'll pick essentially some arbitrary colliders to cite, but the user might prefer to disable ones higher up in the hierarchy instead.

Editor/GlobalColliderPass.cs
bdunderscore: Or, for that matter - do we need to preserve the original object or should we just overwrite the ColliderConfig object in the descriptor?

Editor/GlobalColliderPass.cs
```diff
@@ -0,0 +1,260 @@
...
+#endif
+
+using UnityEngine;
+using Object = UnityEngine.Object;
+
+#endregion
+
+namespace nadena.dev.modular_avatar.core.editor
+{
+	internal class GlobalColliderPass : Pass<GlobalColliderPass>
```
bdunderscore: This particular pass is very VRC-specific - I think when we support other platforms they'll just implement a different pass. So I'd just use `VRChatGlobalColliderPass` with a `[RunsOnPlatforms(WellKnownPlatforms.VRChatAvatar30)]`

Editor/Inspector/GlobalColliderEditor.cs
```diff
@@ -0,0 +1,222 @@
...
+namespace nadena.dev.modular_avatar.core.editor
+{
+	[CustomPropertyDrawer(typeof(GlobalCollider))]
+	internal class GlobalColliderColliderDrawer : EnumDrawer<GlobalCollider>
+	{
+		protected override string localizationPrefix => "global_collider.bone";
+
+		protected override Array enumValues => new object[]
+		{
+#if MA_VRCSDK3_AVATARS
```
bdunderscore: The inspector should not depend on the presence or absence of VRCSDK

Editor/Inspector/GlobalColliderEditor.cs
```diff
@@ -0,0 +1,222 @@
...
+			prop_position = serializedObject.FindProperty(nameof(ModularAvatarGlobalCollider.position));
+			prop_rotation = serializedObject.FindProperty(nameof(ModularAvatarGlobalCollider.rotation));
+			prop_visualizeGizmo = serializedObject.FindProperty(nameof(ModularAvatarGlobalCollider.visualizeGizmo));
+		}
+
+		protected override void OnInnerInspectorGUI()
+		{
+			serializedObject.Update();
+
+#if MA_VRCSDK3_AVATARS
```
bdunderscore: We should be able to display and edit this field even if VRCSDK is not available (although copy original might not work)

docs~/docs/reference/global-collider.md
```diff
@@ -0,0 +1,27 @@
...
+
+![Global Collider](global-collider.png)
+
+## When should I use it?
+
+You can place this component on any game object and define its shape like a standard capsule or VRChat Physbone collider.
+
+## What does it do?
+
+The functionality of this component varies by platform.
```
bdunderscore: We should call out:
1) Currently only vrchat is supported
2) On VRChat, we steal finger bones, so you might run out of global colliders eventually
3) On other platforms, we may implement this in a different way

And probably document the weird contact-only values?

Runtime/ModularAvatarGlobalCollider.cs
ZenithVal: I set this as default because the default/first GlobalCollider is Head (Which is contact only) I would've changed the order of the enum but I didn't want to break component conversion for ZA Remap Collider -> MA Global Collider. 

A "None" value is possible, would making that the default be reasonable?

Runtime/ModularAvatarGlobalCollider.cs
ZenithVal: Yes. (If you're referring to copyOriginalShape)

Editor/Inspector/GlobalColliderEditor.cs
ZenithVal: The thought process on the array here was to show different lists to different platforms.

Editor/GlobalColliderPass.cs
ZenithVal: Sounds good.
image

Runtime/ModularAvatarGlobalCollider.cs
ZenithVal: It's Y. Should this be something on the actual documentation page or just a comment code side?

Editor/GlobalColliderPass.cs
ZenithVal: Have never used refs before so let me know if the recent pushed approach is better.

Editor/Inspector/GlobalColliderEditor.cs
bdunderscore: Yeah but that means if you e.g. accidentally mess with the value on a non-VRC project, you won't be able to set it back.

Runtime/ModularAvatarGlobalCollider.cs
bdunderscore: Code and doc page probably

Runtime/ModularAvatarGlobalCollider.cs
bdunderscore: An unset value would be reasonable yeah

Editor/Inspector/GlobalColliderEditor.cs
```diff
@@ -38,71 +38,83 @@ internal class GlobalColliderColliderDrawer : EnumDrawer<GlobalCollider>
...
+				{
+					EditorGUILayout.BeginHorizontal(EditorStyles.helpBox);
+					EditorGUILayout.HelpBox(S("hint.global_collider.nowarn_manual_overwrite"), MessageType.Warning);
+					if (GUILayout.Button(G("global_collider.enable_overwrite_warnings"),
+						GUILayout.ExpandHeight(true)))
+					{
+						prop_ignoreOverwriteWarn.boolValue = false;
+					}
+					EditorGUILayout.EndHorizontal();
+				}
```
bdunderscore: huh, why not just a checkbox?

Editor/Inspector/GlobalColliderEditor.cs
```diff
@@ -134,87 +146,38 @@ public void OnSceneGUI()
...
+			var globalRot = my.RootTransform.rotation * my.Rotation;
 			HandlesUtil.DrawWireCapsule(globalPos, globalRot, clampedHeight * scale, clampedRadius * scale);
 		}
 
 #if MA_VRCSDK3_AVATARS
 		private void CopyOriginalCollider(ModularAvatarGlobalCollider my)
 		{
-			var descriptor = my.gameObject.GetComponentInParent<VRCAvatarDescriptor>();
-			if (descriptor == null) return;
+			var desc = my.gameObject.GetComponentInParent<VRCAvatarDescriptor>();
```
bdunderscore: `RuntimeUtil.FindAvatarInParents`

Editor/VRChatGlobalColliderPass.cs
```diff
@@ -0,0 +1,171 @@
...
+					if (targetCollider == GlobalCollider.None)
+					{
+						logAutoRemapsFailed.Add(my.gameObject);
+						continue;
+					}
+				}
+
+				usedColliders[targetCollider] = my.gameObject;
+
+				var desc = ctx.VRChatAvatarDescriptor();
```
bdunderscore, line 151: We'll need to conditionally compile this file to avoid breaking the build when VRCSDK isn't present.

Runtime/ModularAvatarGlobalCollider.cs
```diff
@@ -25,34 +30,148 @@ public enum GlobalCollider
...
+		//Returns a reference to the collider config in the descriptor.
+		//also removes mirroring on related colliders.
+		static internal ref ColliderConfig GetVRChatDescriptorCollider(VRCAvatarDescriptor desc, GlobalCollider collider, bool disableMirroring)
+		{
+			switch (collider)
+			{
+				case GlobalCollider.Head: return ref desc.collider_head;
+				case GlobalCollider.Torso: return ref desc.collider_torso;
+				case GlobalCollider.HandLeft:
+					if (disableMirroring) desc.collider_handL.isMirrored = false;
```
bdunderscore: Can't this just be done by the caller...?

Editor/Inspector/GlobalColliderEditor.cs
ZenithVal: This wouldn't be something we'd want anyone outside of advanced power users to use. It could cause issues, especially if someone distributes a prefab with it.

Runtime/ModularAvatarGlobalCollider.cs
ZenithVal: Oops. It is. `collider_handL` was meant to be `collider_handR`. This disables mirroring on the collider pair.

Editor/Inspector/GlobalColliderEditor.cs
bdunderscore: This won't stop someone from distributing a prefab with this. I think the preferred way to handle this would be a project or avatar-level preference (VRChatSettings option?) in this case.

bdunderscore:
This might be buried in a resolved thread, so: Disabling warnings with a hidden setting is something I want to avoid - you shouldn't need to go digging in debug inspectors to use MA features, even if you are an advanced user. I'd prefer to make this a project-level or avatar-level setting. We could use the VRChatSettings component for this - it already errors out if there's multiple instances of the component, and it's intended for these kinds of avatar-level settings that don't make sense being part of a prefab.

ZenithVal:
Removed the hidden debug toggle in favor of a **Low Priority** toggle. 

The scenario where it was desirable to ignore the collider overlapping warning was when using Global Collider to modularize the configuration of a base's colliders. If that collider prefab was used and then a separate collider prefab was added on top, you'd get a warning for no available automatic colliders (or overlapping manual assignments). Now, when a collider is set to Low Priority we no longer track its usage (in a way that matters) which allows the base collider to be hijacked without fuss.

ZenithVal:
In hindsight ~ I should've asked about that merge conflict. I didn't fully understand what I was changing. Sorry about that.

bdunderscore:
I went ahead and created a bunch of unit tests - and found a couple potential bugs in the process:
- A warning was being generated when overriding a low priority collider with a high priority manual collider
- Multiple auto colliders were overwriting the same base collider over and over
- Auto colliders would not avoid low-priority colliders

Do these changes seem sensible to you? If so I'll brush up the docs and merge - thanks for the contribution!

bdunderscore (approved):

ZenithVal:
Good catches! Looks good.

Now adjusted so that low prio colliders are not yoinkable for automatic remaps until they're the last option.

bdunderscore:
I think this is ready to merge now. Thanks for the contribution!

PR #1799, opened 2025-11-03
fix: Convert Constraints breaks constraints replaced with Replace Object

bdunderscore:


bdunderscore:
@copilot Update all four changelog files appropriately for this change

PR #1831, opened 2025-12-07
test: reorganize tests into generic (all-platform) and VRChat-specific tests

bdunderscore:


bdunderscore:
@copilot Continue refactoring tests which depend on the VRChat SDK (VRCSDK) into their own folder.

You should identify which edit-mode unit tests need to be moved under the GenericTests folder, and which should go under the VRChatTests folder. Some hints to get you started:

1. Look at the failures for the standalone GameCI tests on this PR (#1831).
2. Look at the `PluginDefinition.cs` file to see which features are disabled when the VRCSDK is not present, and therefore where the tests are not expected to pass.
3. Tests involving the unity animation system, parameters, or menus generally require VRCSDK.
4. 

Classify tests based on whether they _should_ work on non-VRCSDK platforms, not whether they _do_ work.

Note that some tests depend on VRCSDK by virtue of their test data prefabs containing components which intentionally only function on VRCSDK, while others depend on VRCSDK simply because they do not yet contain the cross-platform avatar root component. Base your decisions on whether the code looks to have the feature enabled or disabled for non-VRCSDK use cases.

You will be unable to run the tests yourself. Do your best to classify things and submit the PR as-is. When you submit your PR, include a summary of the tests you classified, and why you classified each one.

At the end of your work, you should have the following under UnitTests~:
- `_CommonAssets`
- `_PlayModeTests`
- `_SharedInterfaces`
- `_SharedInterfacesImpl`
- `GenericTests`
- `VRChatTests`
- `TestBase.cs`
and the associated .meta files.

## DO NOT

- DO NOT change the contents of any tests or CI logic; you may only move files around.
  - If you believe a test file contains some tests which depend on VRCSDK, and some tests which do not, notate this in your PR description, and place the test in the VRChat folder for now.
  - If you believe a test file should work without VRCSDK, but requires minor changes, place it in the generic folder. We'll make changes in a subsequent step to keep commit history clean.
  - DO move the .meta files for existing folders and files, as you move the folder or file itself.
- DO NOT create .meta files for any new folders you create. I will handle this for you later.
- DO NOT refactor the `_PlayModeTests`. These tests already are entirely VRChat-dependent.

PR #1835, opened 2025-12-10
Unity 6.3 made the default true,true system under an enum

Toys0125:
Adding this fixes compile in Unity 6.3, I found no other issues with the current build of 1.16.0-alpha.0


bdunderscore:
Thanks! I'll probably merge this in the new year - I'm going to be going on vacation shortly and won't have time to deal with any post-merge issues that come up.

PR #1872, opened 2026-02-01
fix: VisibleHeadAccessory fails when a skinned mesh renderer has a null mesh

bdunderscore:
Closes: #1872


bdunderscore:
@copilot Add a unit test for this case

Issue #573, opened 2023-12-22
[Feature request] Add reverse Blendshape sync


blufish1234:
I think it would be great if there is such a script that can set a blendshape's value to 0 when the controller blendshape has a value of 100, and back to its original value when the controller blendshape goes back to 0. It would be in good use in situations such as certain eye morph options may make blinking strange, so resetting these to 0 would make them normal again, and back to normal values when not blinking. 
![image](https://github.com/bdunderscore/modular-avatar/assets/97332474/22332193-c2ae-4a75-b8a9-5ff1e5510dda)
<img width="1280" alt="image" src="https://github.com/bdunderscore/modular-avatar/assets/97332474/2783026e-f781-4a4a-98b1-be436f99a65b">

huideyeren:
This feature would be helpful to have.
Regarding the avatar “Chocolat-chan” released today by Komado, the `Skirt_Corset_OFF` blendshape of the `Skirt` needs to take a value that is exactly opposite to the `Waist_slim` blendshape of the `Body_base` .

Sayamame-beans:
FYI: If there is no need to realtime sync, we can use MA Shape Changer.

bdunderscore:
Implemented in #1987

PR #1987, opened 2026-05-17
feat: Add an inversion option to support inversion of blendshape values

rnmmr:
### Summary

Adds a per-binding "Invert" option to `BlendshapeSync` that flips blendshape values (0→100, 100→0).

### Motivation

Partially addresses #396 and #573.

### Changes

- Added `Invert` bool field to `BlendshapeBinding` (runtime + serialization)
- Updated `BlendshapeSyncAnimationProcessor` to invert both default weights and animation curves at build time
- Updated `BlendshapeSyncEditor` inspector to display an "Invert" checkbox column
- Added localization strings for `en-US` and `zh-Hans`

### Scope & Limitations

This implementation provides **direct inversion** (0→100, 100→0) only. Proportional scaling and custom range mapping were considered but not implemented due to significantly higher complexity (e.g., arbitrary range input, edge cases with keyframe tangents). If there is demand for more advanced mapping options, those can be built on top of this as a follow-up.

### Testing

- Verified that inverted blendshapes correctly reflect 100−weight in Play Mode
- Verified inspector UI displays and persists the invert toggle


bdunderscore:
I'm not super inclined to add a singular invert option - the inevitable follow-up request (which we've already seen) is for arbitrary curves, and this in turn requires a completely different data model. Once that's implemented, this invert option becomes legacy baggage we have to carry around for format compatibility. You're right that this carries more complexity, of course, but I'm not convinced that the linear inverse actually covers enough use cases here to deserve that baggage.

rnmmr:
Swapped bool Invert to an AnimationCurve for remapping blendshape values (0-1 range). Linear transforms like inversion are exact. Non-linear curves approximate between keyframes since we transform each source keyframe individually - accuracy depends on how many keyframes the source animation has. Default is identity, doesn't break existing behavior.

Editor/BlendshapeSyncAnimationProcessor.cs
```diff
@@ -181,9 +184,33 @@ private void ProcessClip(AnimatorServicesContext asc, VirtualClip clip)
...
+                            var tPlus = Mathf.Clamp01(t + epsilon);
+                            var tMinus = Mathf.Clamp01(t - epsilon);
+                            var valPlus = remapCurve.Evaluate(tPlus) * 100f;
+                            var valMinus = remapCurve.Evaluate(tMinus) * 100f;
+                            var slope = (valPlus - valMinus) / (tPlus - tMinus);
+                            remappedCurve.AddKey(new Keyframe(
+                                key.time, val,
+                                key.inTangent * slope,
+                                key.outTangent * slope
+                            ));
```
bdunderscore, line 210: Can't we use the chain rule to determine the tangents analytically instead?

bdunderscore (commented):
Sorry for the late reply, I was out of town for a bit. I had thought tangents would be tricky to deal with, but I think I'm convinced that this works now. I'll test it a bit and if things look okay will probably merge in a bit (I'll need to push some doc updates, etc onto this branch as well first)

Editor/BlendshapeSyncAnimationProcessor.cs
bdunderscore: Nevermind, that would only work if the control points are the same.

PR #2000, opened 2026-06-12
Fix BlendShape Sync remapping animation generation

kb10uy:
* Remapped tangent should not be scaled by 100 (unlike value clamp)
* [AnimationCurve.AddKey](https://docs.unity3d.com/ScriptReference/AnimationCurve.AddKey.html) will *not* add keyframes as-is.
    - Copying original keys and modify instead


bdunderscore:
Do you have a test case for this that can be made into a unit test?

kb10uy:
The first case I found this was relatively simple (like ReLU(x or -x)), which may be suitable for tests.
I'll add them later.

bdunderscore:
Thanks!

PR #2007, opened 2026-06-21
fix: compile error on Unity 6000.0 (but not .4)

bdunderscore:


bdunderscore:
@copilot Update all four changelog files. This change specifically improves compatibility with Unity 6000.0.

PR #2019, opened 2026-06-27
fix: incorrect bindpose calculations in VHA

bdunderscore:
Rewrite VisibleHeadAccessory to fix some totally busted handling. I'm a bit surprised this worked in the first place...


bdunderscore:
@copilot Add CHANGELOG entries to all four changelog files. Changelog entry should be something like: "Fixed incorrect mesh processing in Visible Head Accessory"

PR #2024, opened 2026-06-28
fix: in-use mask in native memory manager might not be maintained correctly

bdunderscore:


bdunderscore:
@copilot update each changelog file with a suitable description of what user-facing impacts this bug might have caused. use the PR number for each entry

PR #2025, opened 2026-06-28
fix: constant-active reaction rule analysis was wrong for inverted conditions

bdunderscore:


bdunderscore:
@copilot add suitable changelog entries to each of the changelog files for this PR

PR #2028, opened 2026-06-28
fix: multiple small issues

bdunderscore:
- **fix(armature-awase): preserve invariant when freeing last segment**
- **fix(merge-armature): guard null lock controller in OnDisable**
- **fix(bone-proxy): avoid hierarchy-changed subscription churn on cache miss**
- **fix(reactive-objects): report fatal error for out-of-range auto int parameters**
- **docs(changelog): add changelog entries**


bdunderscore:
@copilot replace the #XXXX markers in the changelogs with the actual PR number

PR #2034, opened 2026-06-29
BlendshapeSync RemapCurve: fix for RemapCurve in between original curve value and consistency and UI improvements

anatawa12:
**Description**

This PR fixes a bug that RemapCurve does not map curve as expected.

There are two sub-cases for this bug:
- If an animation curve has a keypoint with exactly the same value as a keypoint on the RemapCurve, its left and right tangents will be averaged. We should calculate the left and right tangents separately.
- If there is no keypoint corresponds to the keypoint on the RemapCurve, we should add KeyPoint to change angle of curve.

I think both of them is significant because both can affect merging two blendshapes into one like BreastBig/Small on body to BreastSize on outfit and splitting single BlendShapes to multiple like BreastSize to BreastBig/Small blendshape.

This PR also introduces several UI and consistency improvmeents:
- Provide rect for Animation Curve editor to deny adding keyframes to out of range. This also helps users to understand RemapCurve uses 0...1 instead of 0...100
- Deny non-linear curves in the RemapCurve. supporting bezier curves in RemapCurve would require >3d bezier curve support in animation curve so I added small logic to normalize curve to not have non-linear segements.

**Screenshots**

on master and the remap curve on the left
Screenshot 2026-06-29 at 23 08 14

With commit 236d0f9fb055d10dd19cc1b874af71d70c75870b
Screenshot 2026-06-30 at 1 18 39

With commit 1032426907c03d1daabb471d25747b07d1f068ee
Screenshot 2026-07-02 at 0 37 16

**Commit description**

- 9ff036ced0a6b868ce58b8071ea2b76723d80355: UI: denies adding keyframe at non (0,1) region
- b354ede0757dcc3932ec40c3ae5e37cf64a614e5: consistency: Deny non-linear curves in the RemapCurve.
- efa32295f498cbe84dc0874e9d97db389559fe54: make curve always have key at 0 and 1. Assuming those to have would help us to implement 
- 79829254215afc694eee2f7e42c1f11e093f73ff: refactor: for future testing.
- 3b73b785926afadebae8aec0631d74cfa013866e: refactor: mapping logic can be complex with full curve support so perf improvement for general case
- 236d0f9fb055d10dd19cc1b874af71d70c75870b: fixes half of the problem


bdunderscore:
Hmm - I was under the impression that the current logic would handle nonlinear curves by computing the numerical derivative at the sampled point(s) and deriving the tangent accordingly. This does break with split tangents, though, as you note. But I'd be okay with limiting this to linear curves for the initial release.

Runtime/ModularAvatarBlendshapeSync.cs
```diff
@@ -99,6 +100,26 @@ protected override void OnValidate()
...
+                {
+                    if (remapCurve.length <= 1) continue;
+                    // We ensure key at time = 0 and time = 1
+                    if (remapCurve[0].time > 0) remapCurve.AddKey(0, remapCurve[0].value);
+                    if (remapCurve[remapCurve.length - 1].time < 1) remapCurve.AddKey(1, remapCurve[remapCurve.length - 1].value);
+                    for (int i = 0; i < remapCurve.length; i++)
+                    {
+                        UnityEditor.AnimationUtility.SetKeyBroken(remapCurve, i, true);
+                        UnityEditor.AnimationUtility.SetKeyLeftTangentMode(remapCurve, i, UnityEditor.AnimationUtility.TangentMode.Linear);
+                        UnityEditor.AnimationUtility.SetKeyRightTangentMode(remapCurve, i, UnityEditor.AnimationUtility.TangentMode.Linear);
```
bdunderscore: Do we need to SetKeyBroken here? Should we inherit the broken-key state of the original curve?

Runtime/ModularAvatarBlendshapeSync.cs
```diff
@@ -13,6 +13,7 @@ public struct BlendshapeBinding
         public AvatarObjectReference ReferenceMesh;
         public string Blendshape;
         public string LocalBlendshape;
+        [ui.Curve(0, 0, 1, 1)]
```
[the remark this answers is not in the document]
anatawa12: It's not correct. We can access `nadena.dev.modular_avatar.ui` as `ui` in any namespaces under  `nadena.dev.modular_avatar`.

Runtime/ModularAvatarBlendshapeSync.cs
anatawa12: SetKeyBroken(true) is Inherited from menu operation.
As far as i know, linear tangent mode will never become Broken = false. Setting curve linear will make Broken = true and any operation sets Broken = false will change both tangent mode.
Additionally, linear tangent is purely determined by sibling key so it will never be linked (non-borken).
To prevent bugs in Unity Editor or other tools, I think we should not produce state that is not possible to produce with unity editor.

https://github.com/Unity-Technologies/UnityCsReference/blob/d253e09acffc9803eb9d7cd77f4768d38860d66e/Editor/Mono/Animation/CurveEditor/CurveMenuManager.cs#L312-L317

Editor/BlendshapeSyncAnimationProcessor.cs
```diff
@@ -183,37 +183,45 @@ private void ProcessClip(AnimatorServicesContext asc, VirtualClip clip)
...
+                var valPlus = remapCurve.Evaluate(tPlus) * 100f;
+                var valMinus = remapCurve.Evaluate(tMinus) * 100f;
+                var slopeLeft = (val - valMinus) / (t - tMinus) / 100;
+                var slopeRight = (valPlus - val) / (tPlus - t) / 100;
+                if (float.IsNaN(slopeLeft)) slopeLeft = 1;
+                if (float.IsNaN(slopeRight)) slopeRight = 1;
+
+                key.value = val;
+                key.inTangent *= key.inTangent > 0 ? slopeLeft : slopeRight;
+                key.outTangent *= key.outTangent < 0 ? slopeLeft : slopeRight;
```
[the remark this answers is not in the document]
anatawa12: incorrect. the left/right tangent refers value space mathematical left / right slope. we should use left slope for decreasing tangent and right slope for increasing tangent.

Runtime/UI/CurveAttribute.cs
```diff
@@ -0,0 +1,20 @@
+using UnityEngine;
+
+namespace nadena.dev.modular_avatar.ui
+{
+    internal class CurveAttribute : PropertyAttribute
```
anatawa12, line 5: @bdunderscore where should I place attributes like this and corresponding editor class? I could not find attributes in modular avatar runtime assembly so I added to UI directory since it's inspector user interface related but just `Runtime/` or another directory can also be good I think.

anatawa12:
I have implemented support for splitting basic bezier / hermite curve.

We still need to support other tangent modes, constant curve with infinite tangent, and splitting bezier curve exactly at the place tangent become infinite (current code converts to constant curve).

Refactoring and writing tests are also in progress in my local.

anatawa12:
I failed manually implementing ClampAuto mode so I have changed the way we handle tangent mode to converting curves to free curves before processing instead of manually computing tangent in our code.
Constant curve (curves with infinite tangent) and splitting at vertical are also now supported.

UnitTests~/VRChatTests/BlendShapeSyncRemapCurveTest.cs
```diff
@@ -0,0 +1,568 @@
...
+                outTangent /= time;
+                var curveCase = new AnimationCurveCase(
+                    $"Random Curve (seed: {Seed}, {i}) ({inTangent:G9}, {outTangent:G9}, {leftWeight:G9}, {rightWeight:G9}, {inValue:G9}, {outValue:G9}, {time:G9})",
+                    new Keyframe(time: 0, value: inValue, inTangent: inTangent, outTangent: outTangent, inWeight: 0, outWeight: leftWeight),
+                    new Keyframe(time: time, value: outValue, inTangent: outTangent, outTangent: inTangent, inWeight: rightWeight, outWeight: 0)
+                );
+
+                yield return curveCase;
+            }
+#endif
```
anatawa12, line 446: I have added fuzz testing but I found the test is flaky because of low precision value returned from `AnimationCurve.Evaluate` so I disabled this test by default.
(In addition 100 * 100 matrix is too huge so when we enable the test we have to reduce number of cases)
It looks the precision problem is came from t-axis bezier processing so limiting fuzz testing to hermite curve might improve flakyness though.

For example, test case `TestRemapCurveIsCorrect(Random Curve (seed: 1969480910, 10),Fuzzing Curve (seed: 1969480910, 8)` and `time = 34.05029`, 
`mappedCurve.Evaluate` returned  `32.74703` but it will be `33.9950927670292` when we compute the value with `BezierSegment`.  about 1.248062767 of error.
`animationCurve.Evaluate(time)` returned `10.96229` but `10.8068297173427` with `BezierSegment` and results remapped valut to `34.44767` with `Evaluate` but `33.9950981140137`.

<details>
<summary>modified test code with `BezierSegment`</summary>

```

        [Test]
        [Combinatorial]
        public void TestRemapCurveIsCorrect(
            [ValueSource(nameof(TestCurves))] AnimationCurveCase aniCurveCase,
            [ValueSource(nameof(TestRemapCases))] AnimationCurveCase remapCase
        )
        {
            // This tests that remap is working as expect by comparing remapped curve and remap
            var animationCurve = aniCurveCase.Curve;
            var remapCurve = new RemapCurve(remapCase.Curve);
            var mappedCurve = BlendshapeSyncAnimationProcessor.MapCurve(animationCurve, remapCurve);
            Assert.That(remapCurve.IsIdentity, Is.False);

            var endTime = mappedCurve[mappedCurve.length - 1].time;

            for (float time = 0; time <= endTime; time += 0.01f)
            {
                var original = animationCurve.Evaluate(time);
                var expectValue = remapCurve.GetPointOnCurve(original).MappedValue;
                var mappedValue = mappedCurve.Evaluate(time);

                // we want to allow +/- 1% of error for value (in 0-100) and time (in seconds)
                // 1% in value 0-100 is 1.
                // 1% in time axis is 0.01 sec, and may result in error for 0.01 sec * tangent in value axis 
                var keyRange = Enumerable.Range(0, mappedCurve.length - 1).TakeWhile(i => mappedCurve[i].time <= time).Last();
                var maxTangent = Mathf.Max(Mathf.Abs(mappedCurve[keyRange].outTangent), Mathf.Abs(mappedCurve[keyRange + 1].inTangent));
                var allowedError = Mathf.Max(Mathf.Max(Mathf.Abs(expectValue) / 100, 1), maxTangent / 100);

                double original2;
                double expect2;
                {
                    var keyRange1 = Enumerable.Range(0, animationCurve.length - 1).TakeWhile(i => animationCurve[i].time <= time).Last();
                    var startKey = animationCurve[keyRange1];
                    var endKey = animationCurve[keyRange1 + 1];

                    const float oneThird = 1.0f / 3;

                    var startTangentWeight = (startKey.weightedMode & WeightedMode.Out) != 0 ? startKey.outWeight : oneThird;
                    var endTangentWeight = (endKey.weightedMode & WeightedMode.In) != 0 ? endKey.inWeight : oneThird;
                    var timeSpan = endKey.time - startKey.time;

                    var timeAxisBezier = new BezierSegment(0, startTangentWeight, endTangentWeight, 1);
                    var valueAxisBezier = new BezierSegment(startKey.value, 
                        startTangentWeight * startKey.outTangent * timeSpan, 
                        endTangentWeight * endKey.inTangent * timeSpan, 
                        endKey.value);
                    var isHermite = (startKey.weightedMode & WeightedMode.Out) == 0 && (endKey.weightedMode & WeightedMode.In) == 0;
                    original2 = valueAxisBezier.Compute(isHermite ? time : timeAxisBezier.Solve(Mathf.InverseLerp(startKey.time, endKey.time, time)).First(t => t is >= 0 and <= 1));
                    expect2 = remapCurve.GetPointOnCurve((float)original2).MappedValue;
                }
                
                double actual2;
                {
                    var keyRange1 = Enumerable.Range(0, mappedCurve.length - 1).TakeWhile(i => mappedCurve[i].time <= time).Last();
                    var startKey = mappedCurve[keyRange1];
                    var endKey = mappedCurve[keyRange1 + 1];

                    const float oneThird = 1.0f / 3;

                    var startTangentWeight = (startKey.weightedMode & WeightedMode.Out) != 0 ? startKey.outWeight : oneThird;
                    var endTangentWeight = (endKey.weightedMode & WeightedMode.In) != 0 ? endKey.inWeight : oneThird;
                    var timeSpan = endKey.time - startKey.time;

                    var timeAxisBezier = new BezierSegment(0, startTangentWeight, endTangentWeight, 1);
                    var valueAxisBezier = new BezierSegment(startKey.value, 
                        startTangentWeight * startKey.outTangent * timeSpan, 
                        endTangentWeight * endKey.inTangent * timeSpan, 
                        endKey.value);
                    var isHermite = (startKey.weightedMode & WeightedMode.Out) == 0 && (endKey.weightedMode & WeightedMode.In) == 0;
                    actual2 = valueAxisBezier.Compute(isHermite ? time : timeAxisBezier.Solve(Mathf.InverseLerp(startKey.time, endKey.time, time)).First(t => t is >= 0 and <= 1));
                }

                Assert.That(mappedValue, Is.EqualTo(expect2).Within(allowedError), $"at {time}, actual = {mappedValue}, actual2 = {actual2}, original = {original}, expect = {expectValue}, original computed ours = {original2}, expect with our original = {expect2}");
            }
        }
```

</details>

Editor/BlendshapeSyncAnimationProcessor.cs
```diff
@@ -183,37 +181,373 @@ private void ProcessClip(AnimatorServicesContext asc, VirtualClip clip)
...
+        /// - We want to use Array.BinarySearch to find the index of the closest remap key or segment. There is no generic way to do this with Animation Curve in C# api.
+        /// - We want to get derivative of the remap curve easily. We often require this for mapping implementation.
+        /// - We want to change the behavior of the remap curve for out-of-range values for compatibility reasons.
+        ///   In previous versions of Modular Avatar without remap curves, any out-of-zero range will be mapped as-is.
+        ///   However, AnimationCurve will repeat or clamp the out-of-range values.
+        ///   We need to preserve previous behavior for remap curves looks like identity, so we need to change the behavior for out-of-range values.
+        /// </summary>
+        internal class RemapCurve
+        {
+            private const int MapScale = 100;
```
anatawa12: I think we should reconsider if we should define remap curve as `[0,1]` or `[0,100]`.

Not all blendshapes will have `[0,100]` (most of them are so though) and non-VRChat platforms may supports them (by disabling clamping) and I think raw value mapping is better representation than multiplying by 100 so it might be better to express as `[0, 100]` curve.
For better support for non-`[0,100]` curve, we can disable clamping value range in editor but I think it's rare enough so clamping provides better experience. We might better to add option to change clamp value range or detect range by computing blendShape  maximum / minimum frame value in the future.

anatawa12 (commented):
I have added the tests for RemapCurve and as far as I tried it looks working well so I think this is ready for review & merge with one important spec discussion proposed.

I hope most of important decision in code are written in comment, but I'm sorry if it's too poor.

Editor/BlendshapeSyncAnimationProcessor.cs
bdunderscore: Using a 0-100 scale makes sense to match the SMR inspector UI.

anatawa12: changed in 4a4b6897e85dcc634d6369d96590163570b6daad

anatawa12:
>  error CS2001: Source file '/github/workspace/Packages/nadena.dev.modular-avatar/Runtime/ReactiveObjects/VertexFilters/IVertexFilterBehavior.cs' could not be found.

Library inconsistency with caches?

bdunderscore:
Probably, I'll give it a kick.

bdunderscore:
> modular_avatar_tests.BlendshapeSyncEditorUpdateTest.NullTargetMesh_ContinuesLoop_RatherThanAbortingEarly
>
> System.NullReferenceException : Object reference not set to an instance of an object

Runtime/UI/CurveAttribute.cs
bdunderscore: This is fine for now I think

Editor/CubicSolver.cs
```diff
@@ -0,0 +1,201 @@
...
+            // If a is essentially 0, fall back to a lower-degree solver (quadratic / linear)
+            if (Math.Abs(a) < Eps)
+            {
+                return SolveQuadraticOrLower(b, c, d);
+            }
+
+            // Normalize to monic form: x^3 + B x^2 + C x + D = 0
+            var B = b / a;
+            var C = c / a;
+            var D = d / a;
```
bdunderscore: These divisions would potentially lead to very large input values under special circumstances (ie - values very close to Eps) or incorrect results if a is legitimately nonzero but below Eps. It's also a bit complex math, with many different paths that all need to be evaluated for numerical stability.

Since we don't need to solve _all_ roots, just those on the curve interval from [0, 1], I've replaced this with a solver that avoids this kind of small-value division by performing binary search for roots on intervals delimited by derivatives of the curve. Please do take a look and provide any comments you might have on the new approach.

Editor/CubicSolver.cs
```diff
@@ -0,0 +1,349 @@
...
+                var candidate = x - value / derivative;
+
+                if (!IsFinite(candidate) || candidate <= left || candidate >= right || candidate == x)
+                    candidate = Midpoint(left, right);
+
+                // There are no representable values remaining strictly inside the bracket.
+                if (candidate == left || candidate == right)
+                    return BetterRootCandidate(a, b, c, d, left, right);
+
+                x = candidate;
```
anatawa12, line 191: I'm sorry for delay in response.
I had been considering whether using the Newton method for splitting would improve the convergence rate, but I think it might actually slow it down.
If we consistently use the bisection method, performing it 52 times will bring (right-left) below MachineEpsilon, ensuring convergence.
I don't have any example case 

Therefore, they might be preferable to using the Newton method, that the bisection method or switching between the Newton method and the bisection method based on the size of the candidate range

https://www.desmos.com/calculator/eosslpi2t0

bdunderscore: Hmm. It's a bit hard to say without a benchmark, but I can see your point. The derivative evaluation might incur a fair bit of overhead, not to mention the following division, and it's not clear the reduction in rounds is worth the investment. At a minimum we'd want to symbolically reduce the derivative to multiplication coefficients ahead of time. I'll poke on benchmarking this a bit and see whether it's a win overall.

bdunderscore: https://gist.github.com/bdunderscore/6559761beeeebbcb82f58e9e7993e8a4
```
Newtonian method
---
Unity 2022.3.22f1; runtime 4.0.30319.42000; Stopwatch frequency 10,000,000 Hz
Case | Iterations/sample | Min ns/op | Median ns/op | Max ns/op
Typical_MonotonicAnimationSegment | 131,072 | 2,623.4 | 2,713.0 | 3,169.9
Typical_OvershootingAnimationSegment | 65,536 | 4,417.5 | 4,530.7 | 5,615.0
Stress_NewtonFallsBackToBisection | 131,072 | 2,748.7 | 2,885.8 | 3,311.0
Stress_RepeatedBisection | 65,536 | 3,766.1 | 3,819.9 | 4,678.1

Midpoint-only
BenchmarkSolveCubicInterval (16.556s)
---
Unity 2022.3.22f1; runtime 4.0.30319.42000; Stopwatch frequency 10,000,000 Hz
Case | Iterations/sample | Min ns/op | Median ns/op | Max ns/op
Typical_MonotonicAnimationSegment | 32,768 | 6,404.5 | 6,439.9 | 6,540.5
Typical_OvershootingAnimationSegment | 16,384 | 11,869.4 | 11,893.7 | 11,981.9
Stress_NewtonFallsBackToBisection | 65,536 | 4,790.9 | 4,849.0 | 5,869.9
Stress_RepeatedBisection | 65,536 | 4,804.7 | 4,842.3 | 4,950.4
```

For simple cases, it looks like this is a significant win, and even for pathological cases it's only ~2x overhead - so I think the Newtonian method is a win overall. Thanks for poking me to check though!

anatawa12: I'm a little worried about not ensuring convergence, but it looks performance benefit is big.
Thank you for checking my conserns

bdunderscore: I also did another run using a shuffled set of curves to try to defeat the branch predictor, with similar results:

```
Newtonian
BenchmarkBranchPredictorHostileCorpus (3.863s)
---
Unity 2022.3.22f1; runtime 4.0.30319.42000; Stopwatch frequency 10,000,000 Hz
Pre-generated shuffled corpus: 16,384 distinct cubics
Case | Curves/sample | Min ns/curve | Median ns/curve | Max ns/curve
BranchPredictorHostileCorpus | 65,536 | 3,449.4 | 3,490.2 | 4,726.3

Midpoint-only
BenchmarkBranchPredictorHostileCorpus (3.905s)
---
Unity 2022.3.22f1; runtime 4.0.30319.42000; Stopwatch frequency 10,000,000 Hz
Pre-generated shuffled corpus: 16,384 distinct cubics
Case | Curves/sample | Min ns/curve | Median ns/curve | Max ns/curve
BranchPredictorHostileCorpus | 32,768 | 7,627.7 | 7,750.8 | 7,832.2
```

https://gist.github.com/bdunderscore/8dcd4cda56070dc55fd11a58327235d7

bdunderscore: Convergence is assured because the newton solver is only run on an interval with no zero-crossings of the derivative, and we reject its candidate if it falls outside of the interval or is unmoving. You're right that it's not _guaranteed_ to be better than the midpoint solver however, and there is a MaxIterations bound to terminate in pathological conditions (though it might return an inaccurate root in these cases). We could consider a fallback which switches to midpoint evaluation when we approach this bound, but I'm not sure how likely this actually is in practice...

Editor/BlendshapeSyncAnimationProcessor.cs
```diff
@@ -183,38 +183,292 @@ private void ProcessClip(AnimatorServicesContext asc, VirtualClip clip)
...
+        {
+            if (remapCurve.IsIdentity)
+                return curve;
+
+            // We expect the curve to be in Free mode.
+            for (var i = 0; i < curve.length; i++)
+            {
+                Debug.Assert(AnimationUtility.GetKeyLeftTangentMode(curve, i) == AnimationUtility.TangentMode.Free);
+                Debug.Assert(AnimationUtility.GetKeyRightTangentMode(curve, i) == AnimationUtility.TangentMode.Free);
+            }
```
anatawa12, line 224: I have found that applying `NormalizeCurveToFreeTangents` even for identity `remapCurve` would break compatibility with tools that checks if two animation curves are same or not like https://github.com/anatawa12/AvatarOptimizer/issues/1777.

AvatarOptimizer will be fixed soon, but I want to ask if modular avatar should behaves as before for identity curve.
normalizing on external tools might have risk of floating point error, but in most case normalizing on external tool work I hope.

Here is the small patch that restores original behavior. This patch will also remove cloning `AnimationCurve` if `remapCurve` is identity.

```suggestion
                    clip.SetFloatCurve(dst.ToEditorCurveBinding(asc), MapCurve(curve, remapCurve));
                }
            }
        }

        internal static AnimationCurve NormalizeCurveToFreeTangents(AnimationCurve curve)
        {
            var newCurve = new AnimationCurve(curve.keys)
            {
                preWrapMode = curve.preWrapMode,
                postWrapMode = curve.postWrapMode
            };

            for (var i = 0; i < newCurve.length; i++)
            {
                AnimationUtility.SetKeyBroken(newCurve, i, true);
                AnimationUtility.SetKeyLeftTangentMode(newCurve, i, AnimationUtility.TangentMode.Free);
                AnimationUtility.SetKeyRightTangentMode(newCurve, i, AnimationUtility.TangentMode.Free);
            }

            return newCurve;
        }

        const float epsilon = 1 / 500f;

        internal static AnimationCurve MapCurve(AnimationCurve curve, RemapCurve remapCurve)
        {
            if (remapCurve.IsIdentity)
                return curve;

            curve = NormalizeCurveToFreeTangents(curve);
```

bdunderscore: Can you submit this as a PR?

anatawa12: yes. sure

PR #2037, opened 2026-07-02
Add Unity 6000 compatibility

Toys0125:
# Reasons
- Unity 6000.5 now requires you to use the EntityID system.
- Anything using the InstanceID system no longer compiles in Unity 6000.5 for depreciation error.
# Notes
Did not want 1000 compiler flags, so went and add using alias, for things that change with the new version where possible.

### DIFF
--- a/Editor/CreateBlendTree.cs
+++ b/Editor/CreateBlendTree.cs
@@ -12,17 +12,28 @@ namespace nadena.dev.modular_avatar.core.editor
         static void CreateNewBlendTree()
         {
             ProjectWindowUtil.StartNameEditingIfProjectWindowExists(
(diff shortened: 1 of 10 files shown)


bdunderscore:
There's significant overlap between this and https://github.com/bdunderscore/modular-avatar/pull/2015, I'll take a look this weekend to see how we want to go here.

bdunderscore:
(I also need to set up CI for Unity 6000+...)

bdunderscore:
See also https://github.com/bdunderscore/ndmf/pull/808

bdunderscore:
Merged #2015 instead.

PR #2090, opened 2026-08-13
feat: VRCSDK 3.7.0 compile-time compatibility

anatawa12:
I found there is compile error with SDK 3.7.0 when I'm testing my tool.

I don't know the version support policy in modular avatar, but I found several scripting defines for VRCSDK 3.7.0 or earlier so I created pull request

### DIFF
--- a/Editor/WorldScaleObjectPass.cs
+++ b/Editor/WorldScaleObjectPass.cs
@@ -29,7 +29,7 @@ namespace nadena.dev.modular_avatar.core.editor
                 {
 #if MA_VRCSDK3_AVATARS
                     var c = target.gameObject.AddComponent<VRCScaleConstraint>();
-                    c.Sources.Add(new VRCConstraintSource(fixedPrefab.transform, 1));
+                    c.Sources.Add(new VRCConstraintSource(fixedPrefab.transform, 1, Vector3.zero, Vector3.zero));
                     c.Locked = true;
                     c.IsActive = true;
 #else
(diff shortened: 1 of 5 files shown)


Editor/WorldScaleObjectPass.cs
```diff
@@ -29,7 +29,7 @@ protected override void Execute(ndmf.BuildContext context)
                 {
 #if MA_VRCSDK3_AVATARS
```
anatawa12, line 30: Actually this #if should use `MA_VRCSDK3_AVATARS_3_7_0_OR_NEWER`

```suggestion
#if MA_VRCSDK3_AVATARS_3_7_0_OR_NEWER
```

Editor/WorldScaleObjectPass.cs
```diff
@@ -29,7 +29,7 @@ protected override void Execute(ndmf.BuildContext context)
...
                     var c = target.gameObject.AddComponent<VRCScaleConstraint>();
-                    c.Sources.Add(new VRCConstraintSource(fixedPrefab.transform, 1));
+                    c.Sources.Add(new VRCConstraintSource(fixedPrefab.transform, 1, Vector3.zero, Vector3.zero));
```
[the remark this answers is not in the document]
anatawa12: そもそも3.7.0以前ではVRCConstraintがないって話は上記の通り

Editor/WorldScaleObjectPass.cs
bdunderscore: さすがにもうサポート対象外ですね。
Unity 6のVRCSDKが来たら3.10.x移行に変えます。

